PR #15, opened 2020-01-28
Add events page to display calendar for labs

chungthuang:
Screenshot 2020-01-28 22 29 53

### DIFF
--- a/src/routes/EventsRoute.js
+++ b/src/routes/EventsRoute.js
@@ -1,0 +1,21 @@
+import React from "react";
+
+import Header from "../components/modules/Header";
+import PageBody from "../components/common/PageBody";
+import Footer from "../components/modules/Footer";
+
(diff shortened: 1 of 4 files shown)


src/routes/EventsRoute.js
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+
+import Header from "../components/modules/Header";
+import PageBody from "../components/common/PageBody";
+import Footer from "../components/modules/Footer";
+
+export default class EventsRoute extends React.Component {
+    constructor() {
+        super();
```
third774: If the only thing in the `constructor` is a call to `super()`, you actually just omit the constructor entirely and it will be done automatically!

chungthuang: Thanks! I've omitted the constructor.

Issue #20, opened 2020-01-30
Set up TravisCI to build/deploy 'master' branch to staging endpoint when merged into, and build/deploy 'release' branch to production repo/endpoint when merged into


pyxld-kris:

pyxld-kris:
Created a loose continuous delivery pipeline. We are now able to deploy to a staging area, perform QA, and then release from GitHub.
   - TravisCI now triggers when 'master' branch of this repo is merged into (via pull request) and builds/deploys to our staging environment (gh-pages branch of this repo, which can be accessed at https://staging.devlaunchers.com/).
    - Travis CI now triggers when 'release' branch of this repo is merged into (from master) and builds/deploys into the 'private__website' repo of the 'dev-launchers' organization (can be accessed at https://devlaunchers.com/)

This looks like:
([Pull Request] - perform code review/initial testing) -> ([master branch] - perform QA/overall testing in staging environment) -> ([release branch] -> used by users in production)

Depending on how well this solution serves us, and feedback from contributors, we may instead move to implement a different pipeline and development flow, like the one found here: https://nvie.com/posts/a-successful-git-branching-model/

PR #21, opened 2020-01-30
Remove docs folder

pyxld-kris:
The docs folder was used in an older version to serve the site on GitHub
Pages. We've since moved to a different model to serve our repos (still
using GitHub pages), so this folder is no longer necessary.


pyxld-kris:
Merged, and closing. We've moved to a different routing/deployment setup described in #20

PR #25, opened 2020-02-03
year will now update to current year

sydbails:
added variable for date and year. lmk if this is what you had in mind. - syd

### DIFF
--- a/src/components/modules/Footer/Footer.js
+++ b/src/components/modules/Footer/Footer.js
@@ -10,6 +10,8 @@ export default class Footer extends React.Component {
     super(props);
   }
   render() {
+    const date = new Date();
+    const year = date.getFullYear();
     return (
       <footer className={style.FooterBlock}>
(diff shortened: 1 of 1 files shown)


src/components/modules/Footer/Footer.js
```diff
@@ -37,7 +39,7 @@ export default class Footer extends React.Component {
           </Link>
         </div>
         <div className={style.OrgInfoArea}>
-          Terms of Service | Privacy Policy | ©Dev Launchers, 2020.
+          Terms of Service | Privacy Policy | ©Dev Launchers, {year}.
```
third774: One nit that is _not_ important — I would just inline this:

```js
           Terms of Service | Privacy Policy | ©Dev Launchers, {new Date().getFullYear()}. 
```

sydbails: I can change it - lmk.

pyxld-kris:
Looks good, merging into master!

Issue #27, opened 2020-02-06
{Curriculum Curation} Initial collection of third party learning resources for our students


pyxld-kris:
Also see: #7 (Implementing the UI/UX for the solution described here)

---

## Problem
We don't currently have enough material on the site to keep students active, engaged, and learning for more than a few sessions. This has a some negative side effects
   - Mentors working on advising students in scopes outside of the mentor's core competencies experience friction due to the lack of structure guiding their interactions
   - Students beginning projects are not as prepared as they could be/don't have enough material to revisit key concepts when they encounter issues in their personal projects. 

We also do not currently have the resources to develop the majority of these activities ourselves. 

---

## Solution
We're carefully curating a collection of third party learning resources and activities which are intended to bolster our students' understanding of core concepts and tools. This will keep them engaged on the learning side of our platform for longer periods of time, and provide more conversation starting points and structure for our mentor-student interactions. 

To accomplish this, we're using Google Sheets as a store to collaboratively collect different resources we believe will aid our students. This sheet will connect directly to our frontend, and as new resources are added the site will update live. 

**Dev Launchers currently offers two main tracks for students: Code and Design. We have separate curated collections for each track.**

The sheet is located here: https://docs.google.com/spreadsheets/d/1QV419fM2DHZM59mFK6eYYbYiq6bs4sBUpTwVZ_dZJNg/edit#gid=0

**Please request access if you'd like to add something!**

---

## Guidelines
There are some guidelines and constraints to take into account when adding new activities/resources. These tie into the overall philosophy of our program, technical issues, limits created with the social dynamics of our in-person sessions in mind, and overall user experience concerns. *If you have any thoughts on these please holler in the comments!*

Our Guidelines:
   - **Focus mostly on interactive activities and lessons.** Video and longform supplements are welcome, but hopefully can be kept short and not make up an overwhelming majority of the curriculum
   - **No sign up required, or sign up using a GitHub or Google Account is available.** Click a link, instantly be thrown into a new activity or lesson
   - **Able to be laid out on our site in discrete, bite sized units.** Not only to keep the students used to using our platform as their central hub, but this also gives us more control over their direction and trajectory with the content they're consuming.
   - **Ensure the resources are applicable to what we're building with the students!** We're currently set up with three categories a resource can fall into: General, Games, and Apps. 

---

## Sources
We're working on creating a collection of quality sources we can use to find individual activities and lessons. Please leave any links in the comments!

   - https://phaser.io/examples

---

## The Future
[...]

Issue #7, opened 2020-01-27
Implement UI/UX for categorized collections of external activities and lessons


pyxld-kris:
See also: #27 (Curating Curriculum)

TODO: Write description

pyxld-kris:
Solved, in big part thanks to @zacdemi!

PR #28, opened 2020-02-07
Curriculum curation

pyxld-kris:
This change adds tabbed functionality to our 'Learn' page, with two separate resource collections for Code and Design. It also contains functionality to fetch data from a shared google sheet, where we'll be collecting third party activities and lessons to display on our pages.

This was thrown together quickly in an attempt to get something up for our next weekly session.

See #27  and #7 

(This change is being pushed more quickly than we initially anticipated because of an influx of students last week, which highlighted some weak areas we need to bolster. This pull request will most likely skip review and go directly into staging so we can have it live for tomorrow's session, but please still feel free to review the code and let me know what can be improved!)


pyxld-kris:
Please ignore all those commits prior to February 7th. They're from another branch that is about to be merged into master!

PR #31, opened 2020-02-10
10 struct refactor earn

zacdemi:
Refactored the Earn Route into separate components.


pyxld-kris:
This looks great! Going to leave some notes on specific things as reminders as I'm going through this code, then will create separate issues. Most are from shortcuts that were taken when things were initially thrown together. Please give input wherever!

src/components/modules/Earn/Earn.js
```diff
@@ -0,0 +1,43 @@
...
+import PageBody from "../../../components/common/PageBody";
+
+import SpendCardCollection from "./SpendCardCollection";
+import EarnPoints from "./EarnPoints";
+import CurrentPoints from "./CurrentPoints";
+
+export default function Earn() {
+  return (
+    <PageBody>
+      <br />
```
pyxld-kris: Shouldn't be using <br /> to control page layout. We'll want to use margins and padding to control page formatting

src/components/modules/Earn/SpendCardCollection/SpendCardCollection.js
```diff
@@ -0,0 +1,17 @@
...
+
+// Get content data
+let rewards = require("../../../../content/collections/rewards.json").data;
+
+export default function SpendCardCollection() {
+  return (
+    <div className={styles.collection}>
+      {// rendering our prizes
+      rewards.prizes.map((entry, i) => {
+        return <SpendCard entry={entry} />;
```
pyxld-kris: This might be a nit, but passing in the full `entry` object here feels weird to me. Explicitly declaring the props, or using the spread operator, would make the code a bit more approachable

PR #43, opened 2020-02-20
12 struct refactor learn

### LINKED ISSUE #12: Refactor 'Learn' route into individual components

zacdemi:
Refactored the LearnRoute to separate components. closes #12


src/components/modules/Learn/index.js
```diff
@@ -0,0 +1 @@
+export { default } from ".Learn.js";
```
pyxld-kris: Looks like there's a missing slash between `.` and `Learn.js`

zacdemi: Fixed.

PR #53, opened 2020-02-21
Implement useSheetsContext() in Learn route

### LINKED ISSUE #50: Learn Page: Use data from 'useSheetsContext' hook, instead of fetching data inside of the component itself
Data is now fetched from our Google Sheets store when the app is loaded, and provided to all components in the app through the 'useSheetsContext' hook. We can remove all of the Tabletop and extra state stuff in the Learn component, and use the new hook to access our shared state.

The specific state object we're using for the learn page, inside of the sheets context, is 'learnPageData'. This holds a model in the exact same format as we're currently working with inside of the Learn component.

Depends on #5

pyxld-kris:
Closes #50 

The Learn page now uses the useSheetsContext() hook created in pull request #51 

This has the added benefit of loading the Learn tabs MUCH faster when switching back and forth between pages    🏎(vroom)

### DIFF
--- a/src/components/modules/Learn/Learn.js
+++ b/src/components/modules/Learn/Learn.js
@@ -1,77 +1,14 @@
 import React from "react";
 import "react-tabs/style/react-tabs.css"; // import react-tabs styles
+import { Tabs, Tab, TabList, TabPanel } from "react-tabs";
+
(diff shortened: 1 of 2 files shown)


zacdemi:
Looks great! Wow, tab loading is much faster.

PR #65, opened 2020-03-04
13 struct refactor play

zacdemi:
Refactored Play route into separate components.


src/components/modules/Play/Play.js
```diff
@@ -0,0 +1,112 @@
...
+      let devKeys = Object.keys(devData);
+      let prodKeys = Object.keys(prodData);
+
+      devKeys.forEach(devKey => {
+        let repoEntry = devData[devKey];
+        // search look for substring and return its position of its first occurance
+        if (repoEntry.name.search("project__") == 0) {
+          projectsData.push({
+            title: repoEntry.name,
+            description: repoEntry.description + " | " + repoEntry.html_url,
```
dev-launchers-sandbox: Could we get the URL to be a clickable link on a separate line, maybe a tag with the inner value of "GitHub Repo" or "View Source On GitHub" or something of the sort?

zacdemi: For Sure. I can conditionally render it in the EntryCard if its prop size is 'large' (until we get attachments). Unless you think there's a different way. Another alternative is to have the picture and title link directly to the Github page and put the link to run the game at the top of the repository.

zacdemi: Sure. I created a temporary github attachment image to be include in the header.

Issue #127, opened 2020-06-02
[Members Page] Details On Tiered Mentorship Progression Program


pyxld-kris:
We'd like to showcase our new progression system to potential members!

The entire system is outlined at: https://docs.google.com/spreadsheets/d/1h_jIUfQehq4tVdaL2luNoGlwzo7q9PhZ89rRyNTIzbU/edit?usp=sharing

A mockup of what the body of the page might look like, with minimized and maximized sections, is here: 
https://www.figma.com/file/DfzlDb6zU5uP21khNlneju/DL-Mentorship-Tiers?node-id=0%3A1

This would exist at a new route, devlaunchers.com/#/members

pyxld-kris:
Closed with #130

PR #128, opened 2020-06-10
add mentorship program page

weichen6626:
#127


pyxld-kris:
I love this, thank you so much Ben! I spent a little bit of time and added a placeholder description, with some dummy "Join" buttons that still need to be positioned properly.

I also changed the code structure so that all of the data on the tiers is stored in an object, and then used .map() in the JSX. This should hopefully make doing structural changes to each of those AccordionItems easier so we can play with some style stuff next!

Would love to know what you think

weichen6626:
Thanks, Kris! Both the code and the webpage look a lot nicer now👍👍👍

PR #139, opened 2020-07-04
Replace stories with members in header, footer and hamburger menu

chungthuang:


src/components/common/Header/Header.js
```diff
@@ -53,10 +53,10 @@ export default function Header(props) {
           {authUser ? (
             <img className={style.UserProfilePic} src={authUser.photoURL} />
           ) : (
-            <Button fontSize="1.2rem" onClick={doAuthRequest}>
-              Sign In
+              <Button fontSize="1.2rem" onClick={doAuthRequest}>
```
weichen6626, line 56: Extra space here

PR #141, opened 2020-07-16
Implement Check it page issue#103

elitejam:
https://github.com/dev-launchers/platform__website/issues/103

Implement Check-it page for the issue linked above.
image

I could not figure out how to change the devlaunchers logo color black with the hover of yellow. If anyone could help me with that, that would be nice. Also, feel free to request any changes to the text. I just made it up.


pyxld-kris:
This is awesome, thank you Ben! It looks like you nailed the route and module structure as well 🚀 

## On black logos with yellow hovers
As far as the logo question, I believe it was done using CSS which you can find in the Header component. 

The file is located at src/components/common/Header/Header.module.css

The section that turns the logo yellow starts on line 30:
```
.LogoWrapper:hover {
  filter: brightness(0) saturate(100%) invert(88%) sepia(10%) saturate(6102%)
    hue-rotate(358deg) brightness(101%) contrast(103%);
}
```

You could use a similar CSS filter to create a black logo!

## On the text

I like what you did a lot here, especially with the bolded verbs in the beginning showcasing clear calls to action! I only see one small error in the **Collaborate** section, the sentence doesn't read correctly to me.

## An extra touch

What do you think about throwing a nice image in there to spice things up? We could put something on the left hand side there to break up that big chunk of white, and add a bit of excitement to the page.

elitejam:
Thanks for the feedback :) That makes sense on the collaborate line. I have also fixed the devlauncher logo color based on your explanation.

Here is what it looks like now
image

I just picked an image on the site. We can absolutely use another (just not sure what counts as fair use and all that).

pyxld-kris:
What do you think about using this image for the page? https://ccsearch.creativecommons.org/photos/aa6f4e75-1a91-4cee-a91c-36440a9a90a7

You know, like... check.. it... ♟️  😆 

Everything else looks great!

___

As far as linking to this page, the initial purpose of this page was to have a link that we could put on different social media accounts that then directs people to all of our other social platforms. I do think it would be cool to have an entry point from somewhere on the site though, do you have any ideas?

elitejam:
Whoops accidentally closed the PR. I reopened it :P 

I like the chess piece. It's a good idea. I added the picture in and have pushed up the change.
image

I think we could do several things to link the page into the site.

1. Put it as part of the reworked header as a part of the dropdown menu in "More"

2. Combine it with the join us page because they serve very similar purposes.
Perhaps rather than showing the whole menu just show the upcoming events 

We could take cues from the way Meetup does it?
image

These would both be involved so I can open up another issue depending on which route we want to take.

pyxld-kris:
We spoke about this on the phone, but I love the idea of merging this with the Join page. I think it would be good as a separate issue, so we can push this live! Merging it in now!

PR #243, opened 2020-12-17
Fix links

NateD247:
In this branch, I have changed all the links on the homepage cards to have no hashtags on the hrefs.


pyxld-kris (approved):
Totally rad 🛹

PR #247, opened 2020-12-17
Change card and cardgroup object names

NateD247:
- changed "props.data" to "props.cards" in CardGroup
- changed "props.data" to "props.cardData" in Card


pyxld-kris (approved):
👍

PR #275, opened 2021-02-06
[App & Button Component] moving to styled components global styles & refactored Button component

Enjoy2Live:
I had to move fonts fetching from CSS @import to <link> tag since styled components were recommending that to me, also I used styled-normalize to normalize all different default styles in each browser so that styles work the same way on all browsers!

P.S. don't worry much about the new "_document.js" file, it's basically serving the required styles for the first render within the HTML and then load the rest in the client.

I added more commits and now the Button Component is just a styled one since there's no logic there, also it's worth mentioning that the Button is an <a> DOM tag so if you need to use the <button> DOM tag add `as="button"` as a prop.


NateD247 (approved):
Incredibly sorry for the late update, but this is AMAZING! Everything checks out and you made the code for the buttons very clean! Great job! 😄

PR #294, opened 2021-03-02
[App] moved common components to styled

Enjoy2Live:


src/components/common/Card/StyledCard.js
```diff
@@ -0,0 +1,81 @@
...
+      justify-content: space-between;
+    `}
+  && {
+    flex-direction: ${({ flexDirection }) =>
+      flexDirection ? flexDirection : "column"};
+  }
+  align-items: center;
+  text-align: ${({ textAlignment }) =>
+    textAlignment ? textAlignment : "left"};
+`;
```
NateD247, line 34: Something here is a little off when ran on a local machine. On the home page, large card should have a flex direction of row, but it turns to flex direction of column. I am assuming it has something to do with line 27, but I am unsure. When you have time, run this on your local machine and take a look at the large card that says "Development Labs."

Enjoy2Live:
> Overall, most of the changes check out! Many thanks for refactoring everything towards styled components. Please take a quick look at my comment on lines 16 through 34 about the large card. Thank you so much!

My bad for not checking all the instances of the card. it should be an easy fix!

Enjoy2Live:
@NateD247 apparently you guys weren't adding the needed prop to the organization card so the styled components code was working just fine btw.

NateD247 (approved):
Thank you so much for refactoring the base and getting our large card layouts to display perfectly! 😄

PR #306, opened 2021-03-13
Use semantic versioning to tag releases

chungthuang:
To create new release, merge master branch to release branch. It will trigger the github action defined in `.github/workflows/release.yaml` to cut a release. When there is a new release, `.github/workflows/docker-build-release.yaml` is triggered to build an image tagged with the version because of this step
```
- name: Docker meta
        id: docker_meta
        uses: crazy-max/ghaction-docker-meta@v1
        with:
          images: devlaunchers/platform-website # list of Docker images to use as base name for tags
          tag-semver: |
            {{version}}
```
This solves the staging vs. production separation because `production/flux-patch.yaml` adds a rule to only deploy new image that has a semantic versioned tag.

### DIFF
--- a/.github/workflows/docker-build-release.yaml
+++ b/.github/workflows/docker-build-release.yaml
@@ -1,22 +1,25 @@
 name: release-publish
 on:
-  # Only trigger if the pull request is closed.
-  pull_request:
-    types: [ closed ]
+  release:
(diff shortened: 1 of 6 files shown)


pyxld-kris (approved):
Looks awesome, thank you!

PR #307, opened 2021-03-15
[User Profile] Added user card

Enjoy2Live:
we have to add name and username properties to the context. I realized we don't have these two!


NateD247 (approved):
This is an incredible addition to the page! Also, you look pretty good in your profile picture 😆

PR #308, opened 2021-03-16
[User Profile] added the Points component

### LINKED ISSUE #299: [User Profile] Points component

Enjoy2Live:


NateD247 (approved):
Thank you SO much! This is looking exactly like we imagined it would!

PR #310, opened 2021-03-16
[SignUp] added communication between page and api (username only)

Enjoy2Live:


NateD247 (approved):
Thank you for getting the flow created!! 😸

PR #318, opened 2021-03-27
Added meta tags to improve seo and for making it look better when sharing. Added link previews

hefeliscatus:
**Make sure to check if it works, i wrote this like in less than a minute, so check it before deploying. Dm me on discord if needed . felis catus#7633**

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -1,6 +1,8 @@
 import React from "react";
 //import style from "./HomeRoute.module.css";
+import Head from 'next/head'
 
+      
 import Header from "../components/common/Header";
 import Home from "../components/modules/Home";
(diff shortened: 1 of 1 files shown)


src/pages/index.js
```diff
@@ -11,7 +13,14 @@ export default function Index() {
   }, []);
 
   return (
+    
     <div>
+<Head>
+        <meta property="og:title" content="Dev Launchers" key="title" />
+            <meta property="og:description" content="Launching your imagination, to heights unimaginable" key="title" />
```
NateD247, line 20: I really like this line! 😄

NateD247 (approved):
Thank you so much for creating this update and for keeping the changes small but mighty! Awesome job! 🥳 🎆

PR #320, opened 2021-03-30
fix(chungthuang): Add package config to run semantic release on relea…

chungthuang:
The github action for release wasn't run https://github.com/dev-launchers/platform__website/runs/2197250561?check_suite_focus=true because it wasn't configure to run on release branch, so it defaults to master branch.


NateD247 (approved):
Thank you so much for adding this fix!!

PR #331, opened 2021-04-10
redirecting user to their profile after signing up

Enjoy2Live:
this needs to go production asap so that I can test i﻿t


pyxld-kris (approved):
Thank you

PR #333, opened 2021-04-13
user profile layout

Enjoy2Live:


pyxld-kris (commented):
I really like the bottom area of the page! With the top, can we get a little bit of breathing room between the header and the upper page components?

image

pyxld-kris (approved):
Perfect, thank you!

PR #335, opened 2021-04-14
fix(Enjoy2Live): prevent unauthorized users

Enjoy2Live:
this is for preventing unauthorized users from accessing the user profile :D


pyxld-kris (approved):
This is incredibly helpful, very very very nice!

PR #350, opened 2021-04-17
fix(allsgood1):connected API to BioBox

allsgood1:
I connected the BioBox to our API!


pyxld-kris (approved):
This is so epic! Just tested it and it's working perfectly. 🔥 🔥 🔥

PR #378, opened 2021-05-07
fix(rainbowlotus-games): Make HeaderNav links full height

rainbowlotus-games:
Now the links have more clickable area. Yay.


Enjoy2Live (approved):
LOOOKING GOOOD!

PR #379, opened 2021-05-09
projects  page

Enjoy2Live:
- add Nunito Sans font
- feat(Enjoy2Live): adding projects route
- feat(Enjoy2Live): adding individual projects route


patomation (approved):
yep, theme variables are better than hard coded. Approved.

patomation (approved):
Noice.

pyxld-kris (changes requested):
This is looking great! Visually, there are two things I'd like to try to work in if we can:

1) Some way to *really* draw people to individual projects. Maybe the ability to add images to each project? Or colors, icons, or emojis that make them easy to mentally parse? Here are some of the big reasons people may end up on this page, which should influence the design decisions:
      - **[Developers]** People who might be on the fence about joining our programs or contributing may end up on this page.
      - **[Donors and Companies]** People who are deciding about donating to the organization, or decision makers in companies that may sponsor Dev Launchers
      - **[Members]** With the changes we're making to Dev Launchers, we'll want a page we can send people to so they can find a project with a team they can join
    
2) The space between the two columns is a bit large on my screen. I've attached a screenshot here: 
image

chungthuang (approved):

pyxld-kris (approved):

PR #380, opened 2021-05-09
fix(Enjoy2Live): fix lab times after 12 AM in different timezones

Enjoy2Live:


patomation (approved):
Good, use modules for everything.

pyxld-kris (approved):
Great small PR with a big impact. Thank you!

PR #395, opened 2021-05-24
Implement typewriter effect in home hero section

pyxld-kris:
I've implemented a typewriter effect for our hero section of the home page :smile: It was designed so we can eventually bump it out to a common component and use it in multiple places in the site, and then simply wrap text like so: 

<TypewriterEffect>Some text here</TypewriterEffect>

Right now, though, it is tied directly to our HeroOverlay component. 

https://user-images.githubusercontent.com/46331884/119404327-43ea8200-bca5-11eb-9f37-cd2a8e710a7b.mp4


Enjoy2Live (approved):
LGTM, can't wait to have it as a common component. Thanks!

PR #396, opened 2021-05-25
fix(NateD247): Redesign Support Us Page

NateD247:
- Create new Overlay/Header
- Create new Call To Action


pyxld-kris (approved):
Beautiful 😭

PR #408, opened 2021-05-27
feat(gvsaravanan): Update and add transitions

gvsaravanan:
- Update button transitions
- Update logo transition in header
- Smoother transition for cutout people
- Update scrollbar
- Add transition to partners images
- Update underline transition in footer


NateD247 (approved):
Thank you so much for adding these amazing transitions!!! This will be a great addition to the site!!! 😄

PR #421, opened 2021-05-31
fix(Enjoy2Live): styling issues with the projects page

Enjoy2Live:


NateD247 (approved):
Awesome work!! The styling is on point!! 😃

Issue #403, opened 2021-05-26
[Header, Footer] Remove 'Earn' route from navigation


pyxld-kris:

rainbowlotus-games:
Can be closed.

PR #429, opened 2021-06-02
fix(rainbowlotus-games) Remove earn route from navigation

rainbowlotus-games:
#403


Enjoy2Live (approved):
good job Paulo Dybala!

PR #448, opened 2021-06-12
cleaned up unused packages

Digital-Coder:
we had a lot of old / unused packages so I uninstalled them which also fixed a lot of potential vurnelabilities


Enjoy2Live (approved):
Thanks for uninstalling these useless packages!

Issue #444, opened 2021-06-07
[App] Clicking link to route when already on said route sets page title to that url instead of the name


rainbowlotus-games:
(Sorry for the long title, lol)

**Describe the bug**
When on the home page and clicking the "Dev Launchers" text, next to the logo, it sets the whole title of the page to the url "https://devlaunchers.com/", instead of just staying "Home". Similarly, when on "Create" and clicking "Create" again, it sets the url to "https://devlaunchers.com/create". This happens for all the links in the header.

This also happens on staging.devlaunchers.com (but it sets the title to that url instead).

PR #463, opened 2021-06-16
Nextjs 11 eslint prettier

Digital-Coder:
Main changes;

- removed old packages #448 
- upgraded to Next.js 11 and React 17
- left webpack 4 because of bugs (will try to fix it) (default is webpack 5 in Nextjs 11 now)
- integrated Eslint with eslint-config-next
- added Eslint as recommended plugin for vscode
- enforce default style for everyone by .prettierrc. (default values from documentation)
- added script " lint " in package.json "npm run lint" which will show errors

My main reason to upgrade is because we have all sorts of bugs related to next.js development #444  #353  and to catch those easily and to write code consistently we need Eslint. Next.js 11 supports now it by default which is very easy to set it up. I have been trying to to do it manually for the past couple of days ,but a lot of false positives occurr, some settings are not clear and its a headache.

Their linting is not strong, there are a lot of unused imported packages in .js files which it does not show, but I think its a good start. 

To check what bugs we have Install eslint extension in vscode restart vscode and run "npm run lint"

This commit definately needs testing and let me know if eslint  is linting in console and if prettier enforces code style on save and ofcourse if website builds 
eslint


Digital-Coder:
### Update

- Now on save it should fix some "errors" too. It will sort imports, encourace usage of const vs let
or var, changes concatination to template literals will add noreferrer. These are not "bugs fixes", but maybe nice to have ?
Its always easy to opt out. This airbnb-base config also does not conflict with react / next.

- It will show "error" if there are any unused imports, wrongly imported imports, it will show warnings
if we have console.logs etc.
These pictures are before and after CTRL +S (autosave)
airbnb-addon
airbnb-addon-after-fix

Digital-Coder:
my branch is out of sync with current origin/in-progress and there are too many conflicts. Will close it for now.

PR #469, opened 2021-06-19
Kris/next build error fix

pyxld-kris:
I've temporarily moved things from using getStaticProps() and getStaticPaths() to using a Projects Context which project paths read from and filter through. When we figure out what the 503 issue is with Next.js/Strapi we should definitely move things back for SEO purposes. Temporary fix!


Enjoy2Live (approved):
Thanks Kris, neat solution and hopefully we will get it to work with a better SEO. I added couple of commits to make this work with the project slug being in the url bar :D

PR #484, opened 2021-06-24
Pyxld kris/project touchups

pyxld-kris:


Enjoy2Live (approved):
LGTM, thanks for helping out with the projects page :D

PR #491, opened 2021-06-29
chungthuang: Skip build when container starts, lower staging resource…

chungthuang:
… limit


package.json
```diff
@@ -53,7 +53,9 @@
   "scripts": {
     "dev": "next dev",
     "build": "next build",
-    "start": "next build && next start",
+    "start": "next start",
+    "test": "react-scripts test --env=jsdom",
```
Digital-Coder, line 57: react-scripts does not exist in next.js, needs to be removed

package.json
```diff
@@ -53,7 +53,9 @@
...
+    "eject": "react-scripts eject",
```
Digital-Coder, line 58: react-scripts does not exist in next.js and needs to be removed

chungthuang: Great catch, I accidentally introduced them in a rebase conflict.

PR #508, opened 2021-07-11
fix(alearm246): -update endpoints to strapi -remove discord box since…

alearm246:
… it isn't implemented in strapi yet, so it throws errors


Enjoy2Live (approved):
Thanks!

PR #510, opened 2021-07-13
"Made changes to the icon'

allsgood1:


NateD247 (approved):
Thank you so much for adding this in!!!

PR #553, opened 2021-09-08
Open Roles Component

MarcoB-Ec:
Create Roles component together with modal.


Enjoy2Live (approved):
good job @MarcoB-Ec , The description component looks literally identical to the design and everything works the way expected :fire:

PR #573, opened 2021-11-15
Enjoy2Live/routes for users

Enjoy2Live:
- fix(Enjoy2Live): signing up without spaces
- fix(Enjoy2Live): navbar not linking to user-profile


pyxld-kris (approved):
🦄

PR #599, opened 2022-09-14
feat(recruitment): added recruitment routes to the app

Enjoy2Live:


apps/platform__website/src/components/common/Header/Header.js
```diff
@@ -54,7 +54,7 @@ export default function Header() {
             <NavEntry>SUPPORT US</NavEntry>
           </a>
         </Link>
-        <Link href="/projects">
```
chungthuang, line 57: Does this mean /projects route is removed?

chungthuang (approved):

apps/platform__website/src/components/common/Header/Header.js
Enjoy2Live: It means that the projects routes is not anymore accessible through the links on the footer and header but it's you can still use it with the url.

So now we have a new route called "join" and it's going to serve the recruiting app

PR #633, opened 2022-10-15
fix(RETprojects): added maxlength to UserComment

RETprojects:


RETprojects:
For the last commit I made here this weekend, I also included the Google Tag Manager script and the Hotter script for the IdeaSpace page. I am making sure that they're in the right places.

PR #668, opened 2022-10-18
Jordanmcousins/sorting

jordanmcousins:
-Add sorting dropdown element


apps/ideas/src/components/Card/IdeaCardList.js
```diff
@@ -1,71 +1,102 @@
...
+    {
+      value: 'descending:updated_at',
+      label: 'Recent Ideas',
+    },
+    {
+      value: 'ascending:hourCommitmentMin',
+      label: 'Time Commitment',
+    },
+  ];
+
```
pyxld-kris, line 33: re: sortingConfig structure

Generally, this is a good workable structure, though some improvements could be made to add "safety". With the current implementation, the strings "descending" and "ascending" denote the desired order in which cards will be sorted. The pros of this solution are that it makes working with it pretty fluid, and feels reminiscent of a SQL query. I like that aspect of it. On the flip side, that fluidity leaves it open to some potential confusion for future devs if there are any small spelling mistakes or syntax errors. The danger here is that, in cases like that, there won't be any trace of where the errors are arising if future devs use this component and things don't work how they expect. My recommendation would be to create something to better represent the fact that there is a discrete classification happening with `ascending` vs `descending` so that more verbose errors will be shown when things go wrong.

Some potential solutions I can think of right now include using enum values, using booleans, or creating a struct/class with appropriate attributes to use in place of each array element in `sortingConfigs`. Alternatively, on a __completely__ different path, the same safety could be ensured by keeping things as they are but adding some very explicit error handling inside of the component around this particular string.

apps/ideas/src/components/Card/IdeaCardList.js
```diff
@@ -1,71 +1,102 @@
...
+  const handleSetSortedCards = (sortedCardList) => {
```
pyxld-kris, line 34: I think this function could be removed, and instead `setSortedCards` could be passed directly into the `handleSetSortedCards` prop of the `ProjectFilter` component

apps/ideas/src/components/Card/IdeaCardList.js
```diff
@@ -1,71 +1,102 @@
...
-
-            {cards.map((item) => {
-              return <IdeaCard key={item.id} cards={item} setSelectedCard={setSelectedCard} />
+    <StyledIdeaCards>
+      {loading === true ? (
+        ''
+      ) : (
+        <ProjectFilter
+          sortingConfigs={sortingConfigs}
+          cards={cards}
```
pyxld-kris, line 66: For re-usability, this prop could be renamed to something more general than `cards`. Maybe something like `elements` or `collection`, or something in a similar vein.

apps/ideas/src/components/Card/IdeaCardList.js
```diff
@@ -1,71 +1,102 @@
...
-
-
-
-            {cards.map((item) => {
-              return <IdeaCard key={item.id} cards={item} setSelectedCard={setSelectedCard} />
+    <StyledIdeaCards>
+      {loading === true ? (
+        ''
+      ) : (
+        <ProjectFilter
```
pyxld-kris, line 64: I'm not sure the `Project` part of `ProjectFilter` makes sense for the name of this component

jordanmcousins: I agree, I renamed the ProjectFilter to SortableDropdown to make the actual function of the component more clear and to make the name more general. I updated the imports accordingly.

apps/ideas/src/components/Card/IdeaCardList.js
jordanmcousins: I renamed cards to elements to make it more general.

apps/ideas/src/components/Card/IdeaCardList.js
jordanmcousins: I removed this function and it seems to work fine. I also renamed handleSetSortCards to handleSetSortedElements on the SortableDropdown component to make it more general.

apps/ideas/src/components/Card/IdeaCardList.js
jordanmcousins: Good point Kris. I added the isAscending property back in as part of the sortingConfigs in the IdeaCardList component, and then I am checking to see whether it's true in the SortableDropdown component. Then I am handling the string manipulation on the SortableDropdown component by concatenating the "ascending:" or "descending:" string with the value that is passed in for each option. Now we only need to check the boolean for isAscending to handle this rather than rely on a dev to use the exact string we're looking for in the sortElements function. I updated the comments in the SortableDropdown component to provide some clarity.

PR #700, opened 2022-10-23
feat(UI): move ui library from other repo to this

Enjoy2Live:


pyxld-kris (approved):
Looking good! I changed the `NewUI` to `UI` so we're just working from one folder, and then did a few shims to keep the app functioning while we're in this transitory state. Here's an image that gives a very light summary of said shims. These can be removed when we're further along!
image

PR #722, opened 2022-11-10
feat: icons

holuwateemeeleyin:


Enjoy2Live (approved):
A good start, would be better to have all icons showing up there without importing every single one though but something to worry about later.

Congrats for having your first PR merged with us :rocket:

PR #726, opened 2022-11-15
implemented comment validation with color changes

RETprojects:
Will and I implemented comment validation on the workshopping page by changing the border color of invalid text boxes to red upon submission. The borders are blue when focused on, and alerts are displayed if any invalid input is used.


pyxld-kris:
@RETprojects @liwenlin664477 
I'm not seeing any color changes when testing this locally. The one addition I do see is this extra number that shows min characters, but I don't think I'm seeing anything else on my end.
image

PR #741, opened 2022-11-22
Navigation component

mfvil:


Enjoy2Live:
The hamburger button seems uncentered on mobile.




When clicking on the hamburger, the menu doesn't cover the entire viewport... Is this another thing caused by storybook or something we're missing in the implemention?

Enjoy2Live:
@mfvil I found out while implementing the form for dev recruiters that the hamburger button scrolls down with the page?
image

PR #764, opened 2022-12-15
Checkbox component

Nouri-Anouar:


Enjoy2Live (approved):
I think the issue isn't with left + top properties themselves but the fact that the direct parent of the `Input` component isn't positioned `relative`

anyways I approve this and I'll let ideas team test it!

Issue #753, opened 2022-11-25
[Enhancement] Mobile improvements


JulieMass:

JulieMass:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/IdeaSpace-Team-6350185d9843157427cc900b/issues/dev-launchers/dev-launchers-platform/753?planning-poker) with Zenhub @alicefate @Nouri-Anouar @pyxld-kris @RETprojects

JulieMass:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/IdeaSpace-Team-6350185d9843157427cc900b/issues/dev-launchers/dev-launchers-platform/753?planning-poker) with Zenhub @JulieMass

JulieMass:
Hi @dbradham I'm moving this task back to "review/QA" because even though the navigation in mobile is working well, the sign in button is still not visible in mobile. Can you please give it a look whenever you have a moment? 

![image.png](https://images.zenhubusercontent.com/63400f3d1e79614e45d9ebd9/3e9afdb3-0dfa-4743-a4e5-2ef6ba3c882a)

 ![image.png](https://images.zenhubusercontent.com/63400f3d1e79614e45d9ebd9/6d1d4780-8b8e-47b4-8fe6-5bb2cea4053e)

JulieMass:
Hey @dbradham this is working lovely! Finally we can close it! 🎉 😀

![image.png](https://images.zenhubusercontent.com/63400f3d1e79614e45d9ebd9/17ec38a8-cb28-4aea-8ce3-cab0cdba5490)

![image.png](https://images.zenhubusercontent.com/63400f3d1e79614e45d9ebd9/00b45637-e4d3-4297-af1b-b897f345174d)

PR #785, opened 2022-12-26
Minor Mobile Improvements

pyxld-kris:
This PR includes a bunch of small mobile touchups, as per issue #753

For quick reference:
```
Consideration:
We need to fix the top navigation and the home page for mobile and also the sign in button in the workshopping page isn't visible.

Acceptance criteria:

Top navigation and home page works well for mobile
Sign in button in the workshopping page is visible and works well for web responsive.
```

Before creating this PR, I merged `development/ideaspace` into this feature branch. This should hopefully save you some work Dave!

NOTE: There's an issue with the header in master currently which pushes the page width to be wider than it should. That will make the mobile stuff look less correct, but fixing that is outside the scope of this PR.
image


apps/ideaspace/src/components/common/SignInButton/StyledSignInButton.js
```diff
@@ -3,6 +3,6 @@ import Button from "../Button";
 
 export const MenuButton = styled(Button)`
   @media (orientation: portrait) {
-    display: none;
+    //display: none;
```
dbradham, line 6: Is this comment necessary? Could we delete this line altogether?

apps/ideaspace/src/components/modules/WelcomePage/StyledWelcomePage.js
```diff
@@ -3,9 +3,10 @@ import Link from 'next/link';
 
 export const ButtonArea = styled.div`
   width: 100%;
-  height: 50vh;
+  min-height: 50vh;
+  height: 1px;
   max-width: 1700px;
-  max-height: 500px;
+  //max-height: 500px;
```
dbradham, line 9: instead of commenting, could we just delete this line?

pyxld-kris:
Fixes have been applied, thanks Dave!

dbradham (approved):

PR #790, opened 2022-12-29
666 us as miles hardin i want to click call to actions from the home page that can lead me to the submit my idea page projects page or generate an ideapage

jordanmcousins:


apps/ideas/src/components/modules/IdeaGeneratorButton/IdeaGeneratorButton.js
```diff
@@ -21,26 +21,37 @@ export default function IdeaGeneratorButton(props) {
...
     setModalContent(
       <ModalWrapper>
-        <span style={{fontSize:"1.25em"}}>Loading...</span>
+        <span style={{ fontSize: '1.25em' }}>Loading...</span>
         <br />
         <br />
-        We are querying a "Large Language Model" to generate a new idea using AI. It may take up to 30 seconds!
+        We are querying a "Large Language Model" to generate a new idea using
+        AI. It may take up to 30 seconds!
         <br />
```
dbradham, line 29: Outside the scope of this story, but it would be nice to remove these `<br />` tags

https://stackoverflow.com/questions/1726073/is-it-sometimes-bad-to-use-br

jordanmcousins: I was thinking the same thing. We have one other potential change to make to this modal. Originally, after an idea generated, the Generate Idea button would appear in the modal so that you could click to generate another idea. But since I changed the style of the Generate Idea button a lot, it was obtrusive and didn't look good. If we we want users to have the ability to generate another idea while the modal is still up, we could create a story for that and include removing these <br> tags as part of that story.

apps/ideas/src/components/modules/IdeaspaceNavCard/IdeaspaceNavCard.css
```diff
@@ -0,0 +1,10 @@
+.idea-nav-card {
```
dbradham, line 1: Could we add this as a style to the `StyledWelcomePage` file? I know I helped contribute to creating this file, but using a .css file is not aligned with how we are doing things elsewhere in the project.

apps/ideas/src/components/modules/IdeaspaceNavCard/IdeaspaceNavCard.js
```diff
@@ -0,0 +1,44 @@
...
+}) => {
+  return (
+    <div
+      className="idea-nav-card"
+      style={{
+        background: `linear-gradient(${cardBackgroundColor}, url(${cardBackgroundImage})`,
+      }}
+    >
+      <div
+        style={{
```
dbradham, line 18: Could we also move some of these styles to a new `StyledIdeaspaceNavCard` file?

apps/ideas/src/components/modules/IdeaGeneratorButton/AppIdeaGeneratorButton/AppIdeaGeneratorButton.js
```diff
@@ -24,19 +24,45 @@ let fetchIdea = async (modelParameters) => {
...
+        borderRight: 0,
+      }}
+    >
+      Generate an Idea
+    </div>
+    <div style={{ paddingTop: '9%' }}>
+      Want to generate Ideas for you? Use machine learning to generate an idea
+      to get you started!
+    </div>
+    <div style={{ textAlign: 'right', paddingTop: '1%' }}>
```
dbradham, line 56: Could we add a file, `StyledAppIdeageneratorButton.js` for example, to hold these styles instead of using inline styles?

apps/ideas/src/components/modules/IdeaGeneratorButton/IdeaGeneratorButton.js
```diff
@@ -54,7 +65,7 @@ export default function IdeaGeneratorButton(props) {
         openModal={openModal}
         closeModal={closeModal}
       ></Modal>
-      <div style={{ width: "100%", height: "100%" }} onClick={openIdeaModal}>
+      <div style={{ width: '100%', height: '100%' }} onClick={openIdeaModal}>
```
dbradham, line 68: not required for this ticket, but it would be nice to refactor this styling going forward.

dbradham (approved):
Big improvements from removing inline styles. Going forward we will work to eliminate these from the project where possible for consistency, maintainability, and readability.

Issue #768, opened 2022-12-16
[US - DEV] As MILES HARDIN , I want IdeaSpace’ URL verified in Google search console so that I can  find IdeaSpace as a result when I search organically in search engines for a solution to my problem


JulieMass:
### **Considerations:**
We want to verify IdeaSpace landing page URL prefix in Google search console to learn how we can optimize IdeaSpace’s content for organic search. 

### **Acceptance criteria:**
- landing page (https://devlaunchers.org/ideas or https://devlaunchers.org/ideaspace) is verified in google search console
- IdeaSpace's url google search console is tracking data

![image.png](https://images.zenhubusercontent.com/63400f3d1e79614e45d9ebd9/e41df6e2-c93f-45d4-abf3-342a6db0af22)

JulieMass:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/IdeaSpace-Team-6350185d9843157427cc900b/issues/dev-launchers/dev-launchers-platform/768?planning-poker) with Zenhub @JulieMass @pyxld-kris @dbradham @RETprojects @liwenlin664477 @alicefate @Nouri-Anouar @jordanmcousins

PR #794, opened 2023-01-05
RETprojects/verify-url

RETprojects:
Added URL verification of https://devlaunchers.org/ideas in the Google Search Console by adding a Head section with a meta tag to WelcomePage.js.


apps/app/pages/_app.js
```diff
@@ -52,6 +52,7 @@ function MyApp({ Component, pageProps }) {
           <GlobalStyle />
           <div>
             <Head>
+              <meta name="google-site-verification" content="KUjgcCuL0UXshh3A0F02itHW6KizSyra4BIsFE9Iz8I" />
```
dbradham, line 55: Can we move this to within the `ideas` app? We could put it in the `_app.js` file of our app to achieve the same behavior 

However, since we talked, I got clarification from @JulieMass that this should appear on the Welcome Page only, and not the entire "ideas" app. So we need to go back to your original implementation with that page in /ideas app.

apps/ideas/src/components/modules/WorkshoppingPage/HeartButton.js
```diff
@@ -0,0 +1,17 @@
+// test file for the like button on the workshopping page
```
dbradham, line 1: When you pick up a new ticket, could you please branch off of the `development/ideaspace` branch, to avoid bringing in changes from other tickets?

Since this ticket is currently blocked, we don't actually want to merge the code changes with the changes for #768

PR #811, opened 2023-01-16
Layer, BaseLayer and Overlay modifications

alston-hsu:


Enjoy2Live (approved):
I think the overlay component has to somehow target the HTML tag to cover the whole page, for now I'll merge this

PR #847, opened 2023-02-19
Roman/product cards

mfvil:
Product Cards


.github/workflows/chromatic.yml
```diff
@@ -47,7 +47,8 @@ jobs:
       - name: Discord Channel Notification
         env:
           DISCORD_WEBHOOK: ${{ secrets.UI_BUILDS_WEBHOOK }}
-          DISCORD_EMBEDS: '[{"title": "Storybook build for ${{ github.head_ref }}.${{ github.sha }} ","url":  ${{steps.chromatic.outputs.storybookUrl}}}]'
+          DISCORD_EMBEDS: '[{"title": "UI changes are ready for review: ${{ github.event.pull_request.title }}","description": "Started by ${{ github.event.pull_request.user.login }}. Click [here](${{ steps.chromatic.outputs.reviewUrl }}) to view the Chromatic review.","url": "${{steps.chromatic.outputs.storybookUrl}}"}]'
+
```
Enjoy2Live, line 51: I like the things you added there like who created the PR but I think getting the user who made that commit is better.

the reason the url of chromatic isn't showing is `steps.chromatic.outputs.reviewUrl` is nowhere to be found in the outputs defined in line `19` until `23`, those are the only exposed values given by `chromaui/action@v1` step

Enjoy2Live: `github.event.pull_request.title`
I think getting the title of the PR isn't going to be the case always since builds happen on every push not when PRs are made so I think getting the commit msg is better

mfvil: i tried doing github.event.pull_request.comment but I guess I didnt have a comment in my pr. Im assuming that it works.

Enjoy2Live: I like pull_request.comment idea but the action is triggered on every commit so when members commit something the message on discord won't make sense as the PR is still not up

Enjoy2Live: We can follow your approach here too but only if we change the deployment to happen when a PR is up only, what do you think?

mfvil: I think it would be ideal. Do you want me to implement it?

Enjoy2Live: Are you saying it's ideal to change the flow for builds to trigger when a PR is made or builds trigger on push (currently what we have) and show the commit message in the discord embed?

Enjoy2Live: after checking their docs it seems like triggering the build on `pull-request` event can cause issues 
https://www.chromatic.com/docs/github-actions#recommended-configuration-for-build-events

mfvil: Well that I'm not sure what to do here. I was for doing it on every pull request vs a push. But every push makes sense from a ci standpoint, since there could always be a push on every change. What do you think we should do?

Enjoy2Live: I mean it's great as it is but I prefer that we change this only
https://github.com/dev-launchers/dev-launchers-platform/pull/847#discussion_r1125491158

mfvil: I changed that to comment

Enjoy2Live (approved):

PR #859, opened 2023-03-03
Added typecheck CI

Nouri-Anouar:


.github/workflows/typechecks.yml
```diff
@@ -0,0 +1,28 @@
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16.x'
+      - name: Install dependencies
+        run: yarn
+      - name: Run typecheck tests
+        run: yarn typecheck
+        env:
+          TARGET_URL: '${{ github.event.deployment_status.target_url }}'
```
Enjoy2Live, line 28: type checking doesn't require target url, since it runs without a deployed app. typescript does compilation typechecking and not run time one

Enjoy2Live (changes requested):

.github/workflows/typechecks.yml
```diff
@@ -0,0 +1,28 @@
...
+
+defaults:
+  run:
+    working-directory: packages/UI
+
+jobs:
+  test:
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+    if: github.event.deployment_status.state == 'success'
```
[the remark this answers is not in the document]
Nouri-Anouar: Makes sense. Got it!

.github/workflows/typechecks.yml
Nouri-Anouar: I was not sure about that, I understand now, thank you!

Enjoy2Live (approved):

PR #871, opened 2023-03-05
fix(rashikashaw): resolve issue with next optimized issue for M1 Chips

rashikashaw:
There was an error when I did "yarn install".  `next-optimized-images` uses sharp as it's dependency. [3.0.0-canary.10] version of next-optimized-images does not support the latest version of sharp. Sharp has released a new version because the previous version did not support on apples M1 chip. Since sharp released a new version, next-optimized-image also upgraded to `2.6.2`. So, updated the version and packages got installed without any error.

### DIFF
--- a/apps/app/package.json
+++ b/apps/app/package.json
@@ -21,7 +21,7 @@
     "luxon": "^3.0.1",
     "next": "^12.2.3",
     "next-compose-plugins": "^2.2.1",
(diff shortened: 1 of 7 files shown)


rashikashaw:
> 
Screen Shot 2023-03-07 at 10 50 18 AM

This is the run time error.

Enjoy2Live (approved):

PR #873, opened 2023-03-07
Limit number of roles that appear in Opportunities section to four

sky1122:


pyxld-kris (approved):
Quite an elegant solution. Nice work!

Issue #867, opened 2023-03-05
[DEV] (UI Library) Fix Navigation shared component to allow signed in users to access their profiles by clicking their profile pictures


pyxld-kris:
![image.png](https://images.zenhubusercontent.com/633e1a5b90f8260b0fcb1579/ca3a63ef-77cb-4244-9c7e-b8359a113290)

You can view and inspect the current state of this component at https://www.chromatic.com/component?appId=632ca21aeea1618b7fa0bf48&csfId=molecules-navigation--default-nav&buildNumber=225&k=640369f6738e94f26b5eb75b-1200-interactive-true&h=6&b=-5

To access the code for the component, in the dev-launchers-platform repo/codebase go to the `\packages\UI\src\components\organisms\Navigation` folder

judeamos:
We did not fix this issue, it was done by the universal design team i believe, but this is not an issue anymore.

PR #876, opened 2023-03-08
#867 – Fix navigation

judeamos:
I've used the <Link> component to wrap around the profile  element and link directly to the dev launcher's official website. 
A future story will be created to replace this fix with a dropdown.


pyxld-kris (approved):
LGTM

One thing that will need to be addressed is the hardcoded url "https://devlaunchers.org/users/me". When testing in my local environment, this causes a redirect that bumps you out of the local env and straight to production. This would also happen in staging. That's an issue for another PR though, which I'll be looking at creating soon!

PR #897, opened 2023-03-12
Implement new visual design for dropdown component

jordanmcousins:


apps/ideaspace/src/components/modules/BrowseIdeas/BrowseIdeas.js
```diff
@@ -1,19 +1,19 @@
...
 
 function BrowseIdeas() {
   const [cards, setCards] = React.useState([]);
   const [selectedCard, setSelectedCard] = React.useState({});
   const [loading, setLoading] = React.useState(true);
   const [sortedCards, setSortedCards] = React.useState([]);
-  
-  const defaultOptions = [<option key='0'></option>];
+
+  //const defaultOptions = [<option key="0"></option>];
```
dbradham, line 16: Should this be commented? If it is no longer used, can we delete it?

apps/ideaspace/src/components/modules/BrowseIdeas/BrowseIdeas.js
```diff
@@ -31,6 +31,61 @@ function BrowseIdeas() {
...
+      text: 'Time Commitment',
+    },
+  ];
+
+  const sortElements = (selectedSortCriterion) => {
+    console.log(`selected sort criterion is ${selectedSortCriterion}`);
+    let selectedSortingConfig = sortingConfigs.filter(
+      (configOption) => configOption.label === selectedSortCriterion
+    );
+    console.log(
```
dbradham, line 51: can we remove these console logs?

apps/ideaspace/src/components/modules/BrowseIdeas/BrowseIdeas.js
```diff
@@ -31,6 +31,61 @@ function BrowseIdeas() {
...
+            b[selectedSortingConfig[0].value]
+          ? 1
+          : 0;
+      });
+    }
+    console.log(`cardsClone is ${JSON.stringify(cardsClone, null, 2)}`);
+    setCards(cardsClone);
+  };
+
+  const dummyFunction = (Text) => {
```
dbradham, line 85: can we give this function a more meaningful name?

apps/ideaspace/src/components/modules/BrowseIdeas/BrowseIdeas.js
```diff
@@ -63,22 +118,23 @@ function BrowseIdeas() {
...
         <div>
           <FilterDiv>
-            <label htmlFor="sortBy">Sort By</label>
+            <Dropdown
+              options={options}
+              callbackFn={dummyFunction}
+              title="Sort By"
+              width="lg"
+            />
+            {/* <label htmlFor="sortBy">Sort By</label>
```
dbradham, line 127: can we just delete the old sort by instead of commenting?

packages/UI/src/components/organisms/Dropdown/StyledDropdown.ts
```diff
@@ -43,7 +43,7 @@ export const OptionsContainer = styled.div<DropdownProps>`
 
 export const Options = styled.form<DropdownProps>`
   display: flex;
-  width: 100%;
+  width: 75%;
```
dbradham, line 46: Instead of updating the `width` here, could we pass in `'sm'` or `'md'` as the `width` prop to the `Dropdown`?

This change will affect all instances of Dropdown components which may have some unintended effects.

jordanmcousins: We actually are passing in a width prop to the dropdown. I chose "lg" because there's a lot of available space above the idea cards on the BrowseIdeas page. I added a change to the width of the options themselves because many of the options were extending beyond the dropdown container div. Maybe we could find a different approach to only change the width from 100% under certain conditions?

Enjoy2Live: @jordanmcousins how about we remove this line and remove 
```
label {
    width: 100%;
  }
```
all together

Enjoy2Live (changes requested):

packages/UI/src/components/organisms/Dropdown/StyledDropdown.ts
jordanmcousins: That makes sense, I'll implement that change

dbradham (approved):
Looks great!

PR #900, opened 2023-03-12
fix(alicefate): fix theme, change some styles

alicefate:


apps/ideaspace/src/components/common/IdeaForm/IdeaForm.js
```diff
@@ -42,7 +41,7 @@ const IdeaForm = ({
 	};
 
 	return (
-		<atoms.Box margin='1rem'>
+		<div style={{ margin: '1rem' }}>
```
dbradham, line 44: Sorry if this was confusing, but can we leave this as `atmos.Box`

apps/ideaspace/src/components/common/IdeaForm/IdeaForm.js
```diff
@@ -55,19 +54,18 @@ const IdeaForm = ({
 						style={{
 							maxWidth: '36rem',
 							margin: '1.5rem auto 1.5rem',
```
dbradham, line 56: could we wrap this inside of an `atoms.Box` and pass in these as `props` to that component instead of using inline styles?

apps/ideaspace/src/components/common/IdeaForm/IdeaForm.js
```diff
@@ -55,19 +54,18 @@ const IdeaForm = ({
...
-							borderRadius: 32,
-							backgroundColor: 'white',
 						}}
 					>
 						<AutoSubmitToken />
 						<atoms.Typography type='h4' style={{
 							maxWidth: '37rem',
 							margin: '1rem auto 2.5rem',
 							textAlign: 'left',
+							fontSize: '1.5rem',
```
dbradham, line 65: Is it necessary to change the `fontSize` here? I'm just wondering because this is a component from the design team

PR #922, opened 2023-03-16
User profile/mock user data

HavreLoic:
The image present in the component are placeholders because the image to be used is not yet available in the API or in the double checked I have made in the console this was the best alternative at the moment


HavreLoic:
Oops, Try to merge in the wrong branch

PR #963, opened 2023-03-25
remove styling to make links dark blue after they are clicked

dbradham:
This PR addresses the lack of contrast between the nav links and nav bar after users have visited a link.

Current appearance:
image

Appearance after this change:
image


Enjoy2Live (approved):
LGTM

PR #976, opened 2023-03-31
Task 882 complete

erinsgithubpage:
Changes: 
1. move name from profileCard to bioBox
2. used typography component instead of name
3. changed default user Data context


judeamos (commented):
All looks good

PR #977, opened 2023-04-01
Enable Links to public user profile on the People tab

sky1122:
add router.push('users/${id}') in people to link interesting people to their page.


judeamos:
I've reviewed PR all looks good.

PR #990, opened 2023-04-02
Hredhya/progress bar

Hredhya:
Progress Bar Component

### DIFF
--- a/packages/UI/src/components/molecules/ProgressBar/ProgressBar.tsx
+++ b/packages/UI/src/components/molecules/ProgressBar/ProgressBar.tsx
@@ -1,0 +1,9 @@
+import * as React from 'react';
+import type { ProgressBarProps } from '.';
+import { Wrapper } from './StyledProgressBar';
+
+const ProgressBar: React.FC<ProgressBarProps> = (props) => {
+  return <Wrapper>{props.percentage}%</Wrapper>;
+};
+
+export default ProgressBar;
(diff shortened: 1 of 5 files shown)


packages/UI/src/components/molecules/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,25 @@
...
+} from './StyledProgressBar';
+import type { ProgressBarProps } from '.';
+
+const ProgressBar: React.FC<ProgressBarProps> = (props) => {
+  return (
+    <Wrapper percentage={props.percentage}>
+      <Progress percentage={props.percentage} />
+      <PercentageWrapper>
+        <Percentage percentage={props.percentage}>
+          {props.percentage}%
```
Enjoy2Live, line 18: I think this might be buggy if we pass in a value bigger than 100 or less than 0

Enjoy2Live (changes requested):
Looks very good, I'm really happy to see that you followed our approach of using the design tokens (for example: radiusStyles['radius300']). makes me feel like we built this library in a way that everything is organized. Thanks to our amazing devs!

As I mentioned everything looks neat I just left one small requested change.
keep it up :rocket:

Enjoy2Live:
LGTM but I think we can improve this further by making it more accessible. Let's use `progress` tag instead of endless nested divs

Enjoy2Live (approved):

PR #991, opened 2023-04-03
Havre loic/first time user interests next page 965

HavreLoic:


judeamos:
I've had a look at the code, and the code looks fine. I will merge changes and test it out locally as well. There may be two changes I may need to make after merging:
1. Modifying folder structure so that `<UserOnboarding />` is utilised in the onboarding.js file instead of `<BasicInformation />`.  
     - `<UserOnboarding /> `is a better description of what the page does vs `<BasicInformation /> `

2. I will double-check to see if the context API is being used in the correct _app.js file. 
     - the current change is done in the _app.js located in apps/app path, but what's actually imported and used to construct the pages is the _app.js located in the apps/website path.

I will create a new branch and create a new PR for upcoming changes. 
@HavreLoic Just updating you.

apps/app/pages/_app.js
```diff
@@ -13,6 +13,7 @@ import { ThemeProvider } from 'styled-components';
 import theme from '../styles/theme';
 import Script from 'next/script';
 import iubendaScript from '../scripts/iubendaScript';
+import { OnboardingProvider } from '../../website/src/context/OnboardingContext'
```
judeamos, line 16: I will double-check to see if this is the right _app.js file to import context api

judeamos (approved):

judeamos:
I've had a look at the code, and the code looks fine. I will merge changes and test it out locally as well. There may be two changes I may need to make after merging:
1. Modifying folder structure so that `<UserOnboarding />` is utilised in the onboarding.js file instead of `<BasicInformation />`.  
     - `<UserOnboarding /> `is a better description of what the page does vs `<BasicInformation /> `

2. I will double-check to see if the context API is being used in the correct _app.js file. 
     - the current change is done in the _app.js located in apps/app path, but what's actually imported and used to construct the pages is the _app.js located in the apps/website path.

I will create a new branch and create a new PR for upcoming changes. 
@HavreLoic Just updating you.

PR #1002, opened 2023-04-07
Mohammed/deploying apps

Enjoy2Live:
- feat(ci): deploy apps from their branches
- Update staging-apps.yaml
- added disable Image optimization step
- configure app to be exported
- remove automatic github deployments
- Update wrangler to the latest version
- upgrade wrangler
- bump cloudflare/pages-action to v1

### DIFF
--- a/.github/workflows/staging-apps.yaml
+++ b/.github/workflows/staging-apps.yaml
@@ -49,7 +49,7 @@ jobs:
         run: yarn workspace @devlaunchers/app export
 
       - name: Publish
-        uses: cloudflare/pages-action@1
+        uses: cloudflare/pages-action@v1
         with:
           apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
           accountId: "b0eb1f5510cb71344976cb9cac51e2af"
--- a/.github/workflows/staging-apps.yaml
+++ b/.github/workflows/staging-apps.yaml
@@ -55,3 +55,4 @@ jobs:
           accountId: "b0eb1f5510cb71344976cb9cac51e2af"
           projectName: ${{ steps.extract_branch.outputs.branch }}
           directory: ./apps/app/out
+          branch: main
--- a/.github/workflows/staging-apps.yaml
+++ b/.github/workflows/staging-apps.yaml
@@ -37,7 +37,11 @@ jobs:
 
       - name: Install dependencies 👨🏻‍💻
         run: yarn --frozen-lockfile --silent
-
+        
+      - name: disable Image optimization
+        shell: bash
+        run: "sed -ie 's$unoptimized: false$unoptimized: true$g' ./apps/app/next.config.js"
+        
       - name: Run build
         run: yarn workspace @devlaunchers/app build
 
--- a/.github/workflows/staging-apps.yaml
+++ b/.github/workflows/staging-apps.yaml
@@ -1,4 +1,4 @@
-name: staging-apps
+name: deploy-staging-apps
 
 on:
   push:
@@ -56,3 +56,22 @@ jobs:
           projectName: ${{ steps.extract_branch.outputs.branch }}
           directory: ./apps/app/out
           branch: main
+#I think this needs a bit of explanation.
+#I'm extracting the branch name and modifying it to only show what comes after staging/
+#for example: if this action was triggered on staging/ideaspace the modified string becomes 'ideaspace'.
+
+#I'm doing this because in cloudflare pages dashboard I created these projects
+
+#- 'ideaspace'
+#- 'ideaspace1'
+#- 'ideaspace2'
+#- 'recruitment'
+#- 'recruitment1'
+#- 'recruitment2'
+#- 'projects'
+#- 'projects1'
+#- 'projects2'
+#- 'website'
+#- 'website1'
+#- 'website2'
+#and on every commit that touches the branches of these teams I deploy to the app from the branch name extraction step:  Extract branch name
--- a/.github/workflows/staging-apps.yaml
+++ b/.github/workflows/staging-apps.yaml
@@ -55,4 +55,3 @@ jobs:
           accountId: "b0eb1f5510cb71344976cb9cac51e2af"
           projectName: ${{ steps.extract_branch.outputs.branch }}
           directory: ./apps/app/out
-          gitHubToken: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/staging-apps.yaml
+++ b/.github/workflows/staging-apps.yaml
@@ -1,0 +1,54 @@
+name: staging-apps
+
+on:
+  push:
+    branches:
+      - 'staging/ideaspace'
+      - 'staging/ideaspace[1-2]'
+      - 'staging/recruitment'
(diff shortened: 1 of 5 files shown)


apps/app/package.json
```diff
@@ -6,7 +6,6 @@
   "homepage": "http://devlaunchers.org/",
   "dependencies": {
     "@builder.io/partytown": "0.7.4",
-    "@cloudflare/wrangler": "^1.19.12",
```
Enjoy2Live, line 9: @chungthuang I removed `@cloudflare/wrangler` from the dependencies because it's deprecated and makes my github action fails. is this not used anywhere else?

apps/website/package.json
```diff
@@ -5,7 +5,6 @@
   "keywords": [],
   "homepage": "http://devlaunchers.org/",
   "dependencies": {
-    "@cloudflare/wrangler": "^1.19.12",
```
Enjoy2Live, line 8: @chungthuang I removed `@cloudflare/wrangler` from the dependencies because it's deprecated and makes my github action fails. is this not used anywhere else?

.github/workflows/staging-apps.yaml
```diff
@@ -0,0 +1,58 @@
...
+      deployments: write
+
+    steps:
+      - name: Checkout 🛎
+        uses: actions/checkout@v3
+
+      - name: Extract branch name
+        shell: bash
+        run: echo "branch=$(echo ${GITHUB_REF#refs/heads/} | sed 's/staging\///')" >>$GITHUB_OUTPUT
+        id: extract_branch
```
Enjoy2Live, line 30: I think this needs a bit of explanation.
I'm extracting the branch name and modifying it to only show what comes after `staging/` 
for example: if this action was triggered on `staging/ideaspace` the modified string becomes 'ideaspace'.

I'm doing this because in cloudflare pages dashboard I created these projects
```
- 'ideaspace'
- 'ideaspace1'
- 'ideaspace2'
- 'recruitment'
- 'recruitment1'
- 'recruitment2'
- 'projects'
- 'projects1'
- 'projects2'
- 'website'
- 'website1'
- 'website2'
```
and on every commit that touches the branches of these teams I deploy to the app from the branch name extraction step: ` Extract branch name`

.github/workflows/staging-apps.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Export
+        run: yarn workspace @devlaunchers/app export
+
+      - name: Publish
+        uses: cloudflare/pages-action@v1
+        with:
+          apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          accountId: "b0eb1f5510cb71344976cb9cac51e2af"
+          projectName: ${{ steps.extract_branch.outputs.branch }}
```
Enjoy2Live, line 56: After all of this, your app will be deployed to https://ideaspace.pages.dev for general development or https://ideaspace1.pages.dev and https://ideaspace2.pages.dev if you were doing A/B testing

.github/workflows/staging-apps.yaml
```diff
@@ -0,0 +1,58 @@
...
+  push:
+    branches:
+      - 'staging/ideaspace'
+      - 'staging/ideaspace[1-2]'
+      - 'staging/recruitment'
+      - 'staging/recruitment[1-2]'
+      - 'staging/projects'
+      - 'staging/projects[1-2]'
+      - 'staging/website'
+      - 'staging/website[1-2]'
```
Enjoy2Live, line 13: Pushing to `staging/ideaspace` will trigger a deployment where the link you'll see the results in is https://ideaspace.pages.dev/ , This is your team's staging environment!

For A/B testing you can merge the A testing to `staging/ideaspace1` and B to `staging/ideaspace2` and the url you'll see the results in is https://ideaspace1.pages.dev/ or https://ideaspace2.pages.dev/ depending on the branch you pushed to

Enjoy2Live (commented):
@chungthuang @pyxld-kris @dbradham, would love to hear your thoughts

.github/workflows/staging-apps.yaml
```diff
@@ -0,0 +1,58 @@
+name: staging-apps
```
chungthuang, line 1: Maybe include an action in the name, like `deploy-staging-apps`.

.github/workflows/staging-apps.yaml
chungthuang: Can you add this comment in the file? So it doesn't get lost when the PR is merged.

.github/workflows/staging-apps.yaml
```diff
@@ -0,0 +1,58 @@
...
+      contents: read
+      deployments: write
+
+    steps:
+      - name: Checkout 🛎
+        uses: actions/checkout@v3
+
+      - name: Extract branch name
+        shell: bash
+        run: echo "branch=$(echo ${GITHUB_REF#refs/heads/} | sed 's/staging\///')" >>$GITHUB_OUTPUT
```
chungthuang, line 29: What is `GITHUB_OUTPUT`?

.github/workflows/staging-apps.yaml
```diff
@@ -0,0 +1,58 @@
...
+        id: extract_branch
+        
+      - name: Setup node env 🏗
```
chungthuang, line 32: Do we need the emoji?

apps/app/package.json
chungthuang: It's not used because we don't deploy the website as Cloudflare Workers.

apps/app/package.json
```diff
@@ -43,7 +42,7 @@
     "styled-components": "5.3.5",
     "styled-normalize": "^8.0.7",
     "tabletop": "1.6.3",
-    "wrangler": "^2.0.24"
+    "wrangler": "2.14.0"
```
chungthuang, line 45: We can also remove this.

apps/website/package.json
```diff
@@ -40,7 +39,7 @@
     "styled-components": "5.3.5",
     "styled-normalize": "^8.0.7",
     "tabletop": "1.6.3",
-    "wrangler": "^2.0.24"
+    "wrangler": "2.14.0"
```
chungthuang, line 42: We can also remove this.

.github/workflows/staging-apps.yaml
Enjoy2Live: Not really, I can remove it if you think needed.

apps/app/package.json
Enjoy2Live: I think this is necessary because it makes me interact with cloudflare pages API in an easy way. The other one was conflicting with it since this one is a newer version so I deleted the other one only.

.github/workflows/staging-apps.yaml
Enjoy2Live: It's a way to assign value/s as a result of this step and use it in the next steps, you can see how I used it in line 56 of the same file

.github/workflows/staging-apps.yaml
Enjoy2Live: Will do 😄

PR #1004, opened 2023-04-09
fix(alicefate): fix Warning: Text content did not match. 

alicefate:


Enjoy2Live (approved):
I didn't test this personally on my machine but LGTM

PR #1013, opened 2023-04-12
Refining Useronboarding folder & removing user onboarding context api 

judeamos:
I've restructured the user onboarding folder. 
- It was previously structured like so:  `BasicInfomationPage/UserOnboarding/...files`,
- Now its: `UserOnboarding/..files`.

I also removed the context API that was imported in the global _app.js folder,
- i additionally removed the useronboarding context API from the website folder, its not needed anymore.


judeamos:
Merging changes as this is only folder restructuring.

PR #986, opened 2023-04-01
Story 895, rerouting for public profile

judeamos:
These changes are for user story #895. The changes include: 

1. Creating a new [userId].js file for Next.JS routing. 
2. Removing code that is not needed in the [userId].js file located in `apps/website/src/pages/users/public/[userId].js`. 
3. Deleting old file called [username].js file in two locations: `apps/app/pages/users/[username].js`, `apps/website/src/pages/users/[username].js`. 
4. Updated URL in Peoples component to point to the public URL. Part of #880. 

the new URL path to access public profile is _http://localhost:3000/users/public/[userId]_ . I had to set it up this way to fix the routing issue. 
- I will still be looking for ways to have http://localhost:3000/users/[userId] working. 

This change is linked to user story #880 which @sky1122 (Jingwei) worked on. Once this has been merged, that story will also be completed & ready for testing.

PR #1014, opened 2023-04-12
895 Initial Public Profile – Revamp PR. 

judeamos:
- This is a new pull replacing: #986 PR.
- Build issue on #986 PR has now been fixed in this new PR.
- I was able to get the users/[id] URL path working as well. #986 implementation was users/public/[id]
- Added regex validation to make sure all values passed into url param is a number.


judeamos:
Accidentally chose wrong branch

PR #1015, opened 2023-04-12
895 Initial Public Profile – Revamp PR.

judeamos:
- This is a new pull replacing: #986 PR.
- Build issue on #986 PR has now been fixed in this new PR.
- I was able to get the users/[id] URL path working as well. #986 implementation was users/public/[id]
- Added regex validation to make sure all values passed into url param is a number.


sky1122:
The code looks good to me!

sky1122 (approved):

apps/app/pages/users/[userId].js
```diff
@@ -1,5 +1,6 @@
-import Page from '@devlaunchers/user-profile/src/pages/users/[username]';
+import Page from '@devlaunchers/user-profile/src/pages/users/[userId].js';
 import App from '@devlaunchers/user-profile/src/pages/_app';
+export { getStaticPaths, getStaticProps} from '@devlaunchers/user-profile/src/pages/users/[userId].js'
```
Comment (erinsgithubpage) [inline apps/app/pages/users/[userId].js:4]: Looks good, Jude!

apps/user-profile/src/pages/users/[userId].js
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+import axios from "axios";
+
+import UserProfile from "../../components/modules/UserProfile";
+
+export const getStaticPaths = async () => {
+  return {
+    paths: [{ params: { userId: "/"}}],
+    fallback: "blocking",
```
Comment (Enjoy2Live) [inline apps/user-profile/src/pages/users/[userId].js:9]: @judeamos why do you have empty paths?

Comment (judeamos) [reply apps/user-profile/src/pages/users/[userId].js:9]: I wanted next.js to allow users to enter any value in the `users/[userId]` URL, but I couldn't set userId to be an empty string `userId: ""`, it was throwing a compilation error.  Adding an empty path fixed it. 
 
I've just learnt that i could pass an empty array as paths `paths: []`, so I've updated the code to now pass an empty array, instead of `[{params: {userId: "/"}}]`.

PR #1016, opened 2023-04-13
Refining code & cleaning UserProfile.js file – Part 1

judeamos:
I've made some changes to the UserProfile.js

1. Cleaned up & removed unneeded comments
2. Added comments where needed
3. renamed `otherUser` property to `publicUserData` and reflected where needed
4. Utilised `isPublic` boolean to determine when publicUseData & private user data should be used.
5. Fixed some console errors due to: `<TabPanel key=0>`, removed the "key" property, it was not needed after checking the react-tab doc. 
6. Fixed another console error due to `defaultIndex="0"`, this property does not expect string type, it's expecting number.

### DIFF
--- a/apps/user-profile/src/components/modules/UserProfile/UserProfile.js
+++ b/apps/user-profile/src/components/modules/UserProfile/UserProfile.js
@@ -1,29 +1,35 @@
 import React from "react";
 import { useEffect, useState } from 'react';
-import "react-tabs/style/react-tabs.css"; // import react-tabs styles
+import "react-tabs/style/react-tabs.css";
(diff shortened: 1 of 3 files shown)


apps/user-profile/src/components/modules/UserProfile/UserProfile.js
```diff
@@ -1,29 +1,35 @@
 import React from "react";
 import { useEffect, useState } from 'react';
-import "react-tabs/style/react-tabs.css"; // import react-tabs styles
```
judeamos, line 3: I'm just organising the imports better

HavreLoic (approved):
Everything is still working

PR #1053, opened 2023-04-20
Added stepper with next and back button functionality

nmoon4:
- Added the stepper component
- The pages for the onboarding can be found in apps/user-profile/src/components/modules/UserOnboardingModal/PlatformOnboarding


apps/user-profile/src/components/common/Stepper/Stepper.js
```diff
@@ -0,0 +1,168 @@
...
+    return index >= i ? (
+      <Bar key={i} width={width} />
+    ) : (
+      <EmptyBar key={i} width={width} />
+    );
+  });
+
+  const backButton =
+    // This hides the button if index == 0
+    index == 0 ? (
```
judeamos, line 141: always use strict equality "===" instead of "=="

apps/user-profile/src/components/common/Stepper/Stepper.js
```diff
@@ -0,0 +1,168 @@
...
+    stepsData[index].component
+  );
+  const [activeConfig, setActiveConfig] = useState(stepsData[index].config);
+  const [buttonConfig, setButtonConfig] = useState(
+    stepsData[index].config.buttons
+  );
+
+  const backOnClickHandler = () => {
+    if (
+      (buttonConfig.back.useDefaultOnClick === undefined) |
```
judeamos, line 96: This can be simplified to `if(buttonConfig?.back?.useDefaultOnClick) {...}` if it's undefined it's going to be false.

judeamos: the `?.` is known as Optional Chaining operator, it returns undefined, if the properties does not exist in the object

apps/user-profile/src/components/common/Stepper/Stepper.js
```diff
@@ -0,0 +1,168 @@
...
+        setIndex(index - 1);
+      }
+    } else {
+      buttonConfig.back.onClick();
+    }
+  };
+
+  const nextOnClickHandler = () => {
+    if (
+      (buttonConfig.next.useDefaultOnClick === undefined) |
```
judeamos, line 109: Follow same logic as i've added for `backOnClickHandler`

judeamos (commented):
Two code refinements needed, once done, then can approve

PR #1088, opened 2023-05-02
Havreloic/onboarding modal progress indicator 980

HavreLoic:
- Create Progress Bar reusable component with the few props needed for our UserOnboardingModal component
- Call the ProgressBar component in the UserOnboardingModal component and pass in the props
- Call useUserDataContext(hasDone**SoonToBeImplemented**) to get the update to calculate the percentage
- Implement the percentage function


apps/user-profile/src/components/common/ProgressBar/ProgressBar.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import { useState, useEffect } from "react";
+import ProgressBarStyle from './StyledOnboardingProgressBar';
+
+
+const ProgressBar = ({
+    bgColor,
```
judeamos, line 7: The only property that the progressBar should take at the very moment is `percentage`.

judeamos (commented):
The height, bgWidth, barColor, borderRadius, and bgColor properties are all nice to features to have, making the component very customisable. The only disadvantage is that it's too customisable.  Having to pass all these properties to the component every time someone uses the component takes time. It's better to have presets/skins/themes the user can pick from e.g choosing theme 1, or theme 2. etc

apps/user-profile/src/components/modules/UserOnboardingModal/UserOnboardingModal.js
```diff
@@ -13,6 +15,13 @@ Modal.setAppElement("#__next");
  */
 export default function UserOnboardingModal({ isOpen }) {
     const [modalIsOpen, setModalIsOpen] = useState(true);
+    const { userData } = useUserDataContext;
+
+    let percentage = () => {
+        // hasDone(task) has not yet been created but the implementation will be as follow
+        const task = [userData.hasDonePlatformOnboarding, userData.hasDoneDiscordOnboarding, userData.hasDoneThirdPartyOnboarding];
```
judeamos, line 22: have the const variable `task` outside the percentage function.  You can then use it to dynamically set the "1/4" tasks number e.g `{taskDone/task.length}

apps/user-profile/src/components/modules/UserOnboardingModal/UserOnboardingModal.js
```diff
@@ -13,6 +15,13 @@ Modal.setAppElement("#__next");
...
+        return (taskDone / task.length) * 100
```
judeamos, line 23: where is the `taskDone` coming from?

judeamos (changes requested):

apps/user-profile/src/components/common/ProgressBar/ProgressBar.js
HavreLoic: Okay

HavreLoic: It is but I have also used the property `className`

apps/user-profile/src/components/modules/UserOnboardingModal/UserOnboardingModal.js
HavreLoic: I have commented that part of the code({`taskDone/task.length`} logic) at the end of the line in order to avoid that runtime error

apps/user-profile/src/components/modules/UserOnboardingModal/UserOnboardingModal.js
HavreLoic: I have modified the code

judeamos:
Closing this PR as this feature is not needed for the new UX requirements.

Issue #978, opened 2023-04-01
[Dev](User Profile) Onboarding Modal Creation


ethanzlevin:
User Story:
As Nancy Newbie
I want a modal to display on the user profile page the very first time I view it
So that I can continue to be directed on the onboarding process

Acceptance Criteria
Given I have entered my user info in the "information collection" page
when I am redirected to my personal user profile page for the first time
then a blank modal will be created on top of the user profile page.

ethanzlevin:
https://www.figma.com/file/maDULycPMZpl6t1aRJjsac/Hello-World-Project?node-id=237-2662&t=dAoUPd2pBYfY6rz1-0

judeamos:
This has been completed and is in production.

PR #1093, opened 2023-05-05
Correcting condition for when to render UserOnboardingModal Component #978

judeamos:
I've added condition to only show modal when use is coming from the first information page.
- once user navigates from the onboardinging into the user profile page, a URL query is passed to trigger UserOnboardingModal.js to render.


rinagobo:
I've reviewed your code. I think the code looks good.

HavreLoic (approved):
Looks good and the code is still working

PR #1194, opened 2023-06-18
helioLJ/fix-searchresult-in-projects

helioLJ:
The 'searchResult' const was returning a empty array when typed in searchBar, as a result, no project was shown. So I removed the '?' from keys property in 'options':

```diff
- keys: ['interests?.interest'],
+ keys: ['interests.interest'],
```

with this, projects are now filtered by the 'interests' property.


Devanshusp (approved):
Works as mentioned, thank you for fixing the bug @helioLJ!

PR #1210, opened 2023-06-20
do not render outdated meetings in projects details page

helioLJ:
By removing the `.minus({ days: 6 })` method from the const `now`, it means that only future events will be rendered.

```diff
- const now = DateTime.now().minus({ days: 6 });
+ const now = DateTime.now();
```


Devanshusp:
Works as mentioned, thanks @helioLJ!

Issue #1203, opened 2023-06-19
As a Website Visitor, I do not want to see project milestones listed on the project description page unless I am a member of the project team


pyxld-kris:
---
**TASK**: Hide milestones if user is not logged in and part of the team

You can see an example of a milestone listed at https://devlaunchers.org/projects/user-profile
![image.png](https://images.zenhubusercontent.com/633e1a5b90f8260b0fcb1579/c1e31f30-8e15-40f1-a1aa-97a39a1f91c7)

**ACCEPTANCE CRITERIA**
 - When users are not logged in to the site, or are logged into the site but not part of the team listed on the page they're viewing, the Milestones section is hidden

helioLJ:
Made a P.R. for this issue:

[See here](https://github.com/dev-launchers/dev-launchers-platform/pull/1211/commits/11df3feda40c223fbfdb6cc1a4acb948d3cc0f15)

Devanshusp:
Need a way to test this (need to be able to log in)

PR #1211, opened 2023-06-20
hide milestones for not user and not members of the project team #1203

helioLJ:
I created a function that checks if userId is present in leaders or members of team project and returns a boolean.

```js
function isOnTeam(id, team) {
  const leadersIds = team.leaders.map((leader) => leader.id);
  const membersIds = team.members.map((member) => member.id);

  return leadersIds.includes(id) || membersIds.includes(id);
}
```

Created two constants to hold booleans that check if user id is on the project team and if user is logged.

```js
const checkIfIsOnTeam = isOnTeam(userData.userData.id, project.team);
const isLogged = userData.userData.id === 0 ? false : true
```

So I created a conditional that only renders `Milestones` if user is logged and is part of the team project.

```js
{isLogged && checkIfIsOnTeam ? <Milestones data={project?.board?.ProjectMilestone} /> : null}
```

### DIFF
--- a/apps/site-projects/src/components/modules/Projects/Project/Project.js
+++ b/apps/site-projects/src/components/modules/Projects/Project/Project.js
@@ -23,6 +23,13 @@ import HelpBuild from "./HelpBuild";
 import Sessions from "./Sessions";
 import {useUserDataContext} from "../../../../context/UserDataContext"
 
+function isOnTeam(id, team) {
+  const leadersIds = team.leaders.map((leader) => leader.id);
+  const membersIds = team.members.map((member) => member.id);
+
+  return leadersIds.includes(id) || membersIds.includes(id);
+}
+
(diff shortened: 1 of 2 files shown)


helioLJ:
I perceived there is 2 commits in this branch, cause I forgot to 'git pull' before start this new commit... won't happen again :)

Devanshusp:
Logic makes sense, would like a way to test it (by logging in) @helioLJ
Were you able to test the milestones being displayed (I have not been able to since I can't log in)?

helioLJ:
> Logic makes sense, would like a way to test it (by logging in) @helioLJ Were you able to test the milestones being displayed (I have not been able to since I can't log in)?

Unfortunately I couldn't, I just assumed that if user ID was 0, he wasn't logged

PR #1212, opened 2023-06-22
remove click and hover behavior in project page

helioLJ:
Removed the `<a>` tags that was envolving the images and removed the hover styling that was creating the opacity behavior


Devanshusp (approved):
Works as expected, thank you @helioLJ!

PR #1214, opened 2023-06-22
Yacine/fixing nav lists warning and adding other search keys to projects page

yacineKahlerras:
- fixed "key prop" warning in navigation
- added more keys to the fuse search in the projects page

### DIFF
--- a/packages/UI/src/components/organisms/NavDropdown/NavDropdown.tsx
+++ b/packages/UI/src/components/organisms/NavDropdown/NavDropdown.tsx
@@ -1,5 +1,5 @@
 import Link from 'next/link';
-import { useState, useEffect, useRef } from 'react';
+import React, { useState, useEffect, useRef } from 'react';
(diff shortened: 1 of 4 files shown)


packages/UI/src/components/organisms/NavDropdown/NavDropdown.tsx
```diff
@@ -46,12 +46,12 @@ const NavDropdown = ({ title, isOpen = false, links }: NavDropdownProps) => {
         <Options>
           {links.map(({ text, href }, i) => {
             return (
-              <>
+              <div key={i}>
```
Enjoy2Live, line 49: please don't add a new div, this might change the behavior of the component, do this instead to add key prop:
```
<React.Fragment key={i}>
....
</React.Fragment>
```

yacineKahlerras: good idea thank you !

PR #1217, opened 2023-06-23
added options to the donate button

yacineKahlerras:
when clicking on donate button it takes you to the support-us page
after going to the support-us page after clicking on the donate button you can now go back to the project page using the browser back button


Devanshusp (approved):
Thank you @yacineKahlerras, both the support buttons work and the back button redirects to the correct page!

PR #1218, opened 2023-06-23
353/user data fetched on page load

dbradham:
This pull request addresses 3 things:

1. Removes EnvironmentVariables.js files from the app and standardize usage of environment variables
2. Removes all UserDataContext files except the one in packages
3. Adds a hook for setting user data in local storage to avoid repeated requests


RETprojects (commented):
Aside from the failed Storybook tests, these changes look good to me.

packages/UI/src/context/UserDataContext.js
```diff
@@ -29,39 +27,49 @@ export const DEFAULT_USER = {
...
-          socialMediaLinks: currentUser.profile.socialMediaLinks,
-          totalPoints: currentUser.point.totalPoints,
-          totalSeasonPoints: currentUser.point.totalSeasonPoints,
-          availablePoints: currentUser.point.availablePoints,
-          volunteerHours: currentUser.point.volunteerHours,
-          interests: currentUser.interests,
-        });
-        setIsAuthenticated(true);
+  useEffect(() => {
+    localStorage.setItem('userData', JSON.stringify(userData));
```
Enjoy2Live, line 34: @dbradham 
why are we using localStorage? isn't data persistent on every navigation to a different page?

dbradham: The data is not persistent on every navigation to a different page from my testing. This behavior is confirmed in production. This solution does not address the root issue of files like `Navigation.tsx` losing their state with every page load. This just reduces the amount of requests we are making because of files like that losing their state.


image

Enjoy2Live: This sounds like something that can be fixed with the used npm package (*constate* here).
I've made quick example with the same data set and hook with little modification on it. https://stackblitz.com/edit/stackblitz-starters-ocgwke?ctl=1&embed=1&file=pages%2F_app.js

Taken from constate docs => [advanced example](https://github.com/diegohaz/constate#advanced-example)

dbradham: @Enjoy2Live I'm don't understand what you are recommending to change here. 

We are already using constate in this file, but the state is still being reset. Is there something you see that we could improve with our implementation of constate? (I didn't implement the constate code, but it looks correct to me)

Enjoy2Live: @dbradham Have you looked into the example I provided? it basically fixes the issue of redundant-fetching.
We can standardize the example for teams to use the context this same way the example does

dbradham: @Enjoy2Live Have you tested this within the application? I can't navigate to new pages in your example, so I'm not sure how to test that it fixes the problem here. I'm also not able to implement a change like the example to solve the issue.

Please let me know if you know how I can test this within the application.

Enjoy2Live: I can make a PR that fixes this and point it to the PR we're talking about here, I'm not sure what you mean by I can't navigate pages in my example but there's a link that says go to "test" and the other page says go to "home".

These two separate routes/pages

dbradham: @Enjoy2Live Okay, I see. I replaced the changes you made to the context class with the current context class. I am not able to reproduce the redundant requests. (https://stackblitz.com/edit/stackblitz-starters-ocgwke?ctl=1&embed=1&file=pages%2F_app.js)

I'd be happy for you to make this PR, but are you sure this will fix the unwanted behavior in the app? It seems the test environment is fundamentally different than our hosted solution here.

My understanding is that this problem is due to components which call on the UserDataContext having their state reset, not anything within this file. E.g. Navigation.tsx loses all stateful variables with each page load.

dbradham: To go back to your original question, data does not seem to be persistent with each page load. Nothing we change in the Context will affect this behavior per my understanding of the root issue.

This is a short-term patch Kris proposed, and I do not have any better solution until the root issue of data not being persistent with each page load across the app. I'm happy to revisit this discussion if you have any insights into the root issue here. Leaving the current implementation and behavior as it is is a risk because it adds lots of extra stress to the backend.

If you would like for me to comment out the use of localStorage so we can test in staging, I will. I'd like to get the rest of these changes merged sooner than later because this change is already in our `development/ideaspace` branch so it is blocking us from deploying.

Enjoy2Live: @dbradham here's the PR that's supposed to fix data not being persistent between page navigation
https://github.com/dev-launchers/dev-launchers-platform/pull/1335

Enjoy2Live: do you think we deploy it to https://ideaspace1.devlaunchers.org to test is a good idea?

PR #1220, opened 2023-06-23
adding redirect to the join now link

yacineKahlerras:
both **join now** buttons in the project page will redirect the user to the _**join**_ page


Devanshusp (approved):
Works perfectly, thank you @yacineKahlerras!

Issue #1115, opened 2023-05-13
[Dev](User Profile) Animate the last* onboarding tour page


ethanzlevin:
As Nancy Newbie
I want to view an animated dev launchers logo on the first onboarding tour page
so that I am more engaged

Given I have clicked the 'Platform Onboarding' Task
When the page loads
The Dev Launchers Logo will continuously bounce up and down, and the shadow will grow and shrink alongside it.
& the Next button will remain clickable during the animation

judeamos:
The initial story was for V2 design of the onboarding process. We did not create a new story to cover the change to V4 where the animation was moved to the last page.

PR #1221, opened 2023-06-24
Animation first onboarding page (#1115)

rinagobo:
1. Modified the previous code in the pageOne.js to fix the error on <PageOneContainer/> and image source
2. Added flamer-motion package in to User Profile project
3. Changed the color of the spaceship logo to black
4. Added motion to the spaceship logo to bounce up and down 
5. Added shadow shape and animated it to grow and shrink in horizontal direction

### DIFF
--- a/apps/user-profile/package.json
+++ b/apps/user-profile/package.json
@@ -14,6 +14,7 @@
     "autosave": "^1.0.3",
     "axios": "^0.27.2",
     "constate": "^3.3.2",
+    "framer-motion": "6.2.4",
     "luxon": "^3.0.1",
     "next": "^12.2.3",
     "next-compose-plugins": "^2.2.1",
--- a/apps/user-profile/src/components/modules/UserOnboardingModal/PlatformOnboarding/PageOne/PageOne.js
+++ b/apps/user-profile/src/components/modules/UserOnboardingModal/PlatformOnboarding/PageOne/PageOne.js
@@ -1,5 +1,14 @@
 import { Typography } from "@devlaunchers/components/components/atoms";
-import welcomePageOnbaordingLogo from "./../../../../../images/welcomePageOnboardingLogo.png";import { PageOneContainer } from "./StyledPageOne";
+import "/Users/rinabo/developments/dev-launchers-platform/apps/user-profile/src/images/logo-monogram.png" 
+import LogoMonogram from '../../../../../images/logo-monogram.png'
+import { motion } from "framer-motion";
+import { PageOneContainer } from "./StyledPageOne";
+
+const bounceTransition = {
+    y: {
+        yoyo: Infinity,
+    }
+};
 
 export default function PageOne({name}){
     return (<>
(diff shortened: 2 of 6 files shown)


apps/user-profile/src/components/modules/UserOnboardingModal/PlatformOnboarding/PageOne/PageOne.js
```diff
@@ -1,5 +1,14 @@
 import { Typography } from "@devlaunchers/components/components/atoms";
-import welcomePageOnbaordingLogo from "./../../../../../images/welcomePageOnboardingLogo.png";import { PageOneContainer } from "./StyledPageOne";
+import "/Users/rinabo/developments/dev-launchers-platform/apps/user-profile/src/images/logo-monogram.png" 
+import LogoMonogram from '../../../../../images/logo-monogram.png'
+import { motion } from "framer-motion";
```
Enjoy2Live, line 4: Hi @rinagobo , First of all I want to thank you for making the website more lively by adding animations!
just wanted to let you know that using motion is discouraged if you're only doing domAnimations so please read this https://www.framer.com/motion/guide-reduce-bundle-size/ and let me know what you think :rocket:

apps/user-profile/package.json
```diff
@@ -14,6 +14,7 @@
     "autosave": "^1.0.3",
     "axios": "^0.27.2",
     "constate": "^3.3.2",
+    "framer-motion": "6.2.4",
```
Enjoy2Live, line 17: We use framer-motion version 6.5.1 throughout the monorepo, it's better be consistent with the rest of the apps when it comes to deps versions :)

apps/user-profile/src/components/modules/UserOnboardingModal/PlatformOnboarding/PageOne/PageOne.js
rinagobo: Hi Mohammed, thank you for checking my code. I read through the document you attached. Is the easiest way to minimize the bundle size of my animation is putting <LazyMotion strict></LazyMotion> around my <motion.img /> and <motion.div /> ?

Enjoy2Live: The easiest way is to change the import statement from
`import {motion} from "framer-motion"` to `import {m} from "framer-motion"`

And<motion.div> to <m.div> etc

Then eventually wrap your<m.> Component with <LazyMotion features={domAnimation}>

rinagobo: Thank you for your explanation! 

When do you write code like the one in the bottom of the document (SS attached)? I thought this could work too.

Screenshot 2023-07-09 at 5 22 29 PM

Enjoy2Live: You can do that too and wrap your whole app with lazymotion with strict and features prop passed in, _app.jsx in the pages folder is where you should place lazymotion component around the {children}

apps/user-profile/src/components/modules/UserOnboardingModal/PlatformOnboarding/PageOne/PageOne.js
```diff
@@ -1,7 +1,7 @@
 import { Typography } from "@devlaunchers/components/components/atoms";
 import "/Users/rinabo/developments/dev-launchers-platform/apps/user-profile/src/images/logo-monogram.png"
```
Enjoy2Live, line 2: I think this import causes the app to break, because you're importing things from your device rather than relative paths. 
always importing with relative paths like in line 3

PR #1233, opened 2023-06-27
Alston/percy - Integrating Percy

alston-hsu:


.github/workflows/percy.yml
```diff
@@ -0,0 +1,26 @@
...
+      - name: Node 16.x  
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16.x
+
+      - name: Install @percy/cli and @percy/storybook
+        run: yarn add @percy/cli @percy/storybook
+        
+      - name: Percy Storybook Snapshots
+        run: yarn percy storybook https://632ca21aeea1618b7fa0bf48-axgzbckakp.chromatic.com/
```
Enjoy2Live, line 24: line 24 should not be fixed and should have a dynamic value from the chromatic CI job

Enjoy2Live: having a fixed value of https://632ca21aeea1618b7fa0bf48-axgzbckakp.chromatic.com/ means percy will only take snapshot of that storybook always

alston-hsu: Updated

packages/UI/src/components/atoms/Link/Link.tsx
```diff
@@ -4,7 +4,7 @@ import type { LinkProps } from './index';
 
 const Link: FC<LinkProps> = ({ href, text, ...props }) => {
   return (
-    <StyledLink href={href} {...props}>
+    <StyledLink href={href} {props}>
```
Enjoy2Live, line 7: can you revert this please?

alston-hsu: Yeah done

packages/UI/snapshots.yml
```diff
@@ -0,0 +1,2 @@
+- name: Local Storybook atom snapshots
+  url: http://localhost:6006
```
Enjoy2Live, line 2: what does this file do? where is it taken from?

.github/workflows/percy.yml
```diff
@@ -0,0 +1,26 @@
...
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: Node 16.x  
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16.x
+
+      - name: Install @percy/cli and @percy/storybook
+        run: yarn add @percy/cli @percy/storybook
```
Enjoy2Live, line 21: @percy/cli and @percy/storybook are already installed in package.json why adding them in CI instead of just running yarn install?

.github/workflows/percy.yml
```diff
@@ -0,0 +1,26 @@
...
+
+      - name: Percy Storybook Snapshots
+        run: export scheme=https:// && export gitHash=$(git rev-parse --short HEAD) && export appID=632ca21aeea1618b7fa0bf48 && export chromaticURL=${scheme}${gitHash}--${appID}.chromatic.com
+        env:
+          PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
```
Enjoy2Live, line 26: here you're just exposing environment variables to CI but you're not making use of them, can you explain what you're trying to achieve here?

packages/UI/snapshots.yml
alston-hsu: This is for taking snapshots with Percy through a different way. It was taking from reading Percy's documentation.

.github/workflows/percy.yml
alston-hsu: Its another method versus doing yarn install. Here, since this is using Percy, being more specific about what you are installing gives more information about the context.

.github/workflows/percy.yml
alston-hsu: Percy needs a environment variable defined in order to use it with CI/CD (https://docs.percy.io/docs/environment-variables).

Enjoy2Live: I meant there's nothing using those environment vars but it seems like you resolved this in this [752acff](https://github.com/dev-launchers/dev-launchers-platform/pull/1233/commits/752acffb8d158ef24e79785eb4e49df3120c0bc0)

packages/UI/snapshots.yml
Enjoy2Live: is it still needed? when you say different way does that mean non CI way? if so can we get remove it

.github/workflows/percy.yml
alston-hsu: The problem is that Chromatic is failing and the URL used for Percy snapshots is giving a 404 error now.

packages/UI/snapshots.yml
alston-hsu: You can use it locally or with CI/CD. Either way, I removed it.

.github/workflows/percy.yml
Enjoy2Live: What are the common issues you see chromatic failing at?

Enjoy2Live: What are the common issues you see chromatic failing at?

alston-hsu: There's various errors that come up (TypeError: virtualMocks.get is not a function), with the tests failing. I do think it would be better to have something else for the deployment link, and use Percy that way instead.

Issue #1228, opened 2023-06-25
Cannot close dialog box on project page


ethanzlevin:
**Describe the bug**
On Project Page, users with edit access cannot close out dialog box for editing permissions

**To Reproduce**
1. Go to https://devlaunchers.org/projects/user-profile (or any project page you have admin access for)
2. Click on ![image.png](https://images.zenhubusercontent.com/6404cf78eb446cd49a9a130a/db134cc7-624e-4408-bb8d-39b0cccf6da2)
3. no response

**Expected behavior**
dialoge box should close

**Desktop (please complete the following information):**
 - OS: windows 11
 - Browser chrome
 - Version ??

helioLJ:
@Devanshusp Think I'm gonna need a Admin Permission to test this dialog box.

PR #1249, opened 2023-07-05
fix close dialog box editor notification #1228

helioLJ:
- [x] Created a `State  to check if editor `isClosed` or not, by default it's `false`, since the dialog starts open.
- [x] Created a condition that it will return a `fragment` if `isClosed` it's **true**.
- [x] Created a **function** responsible for close the `EditorNotification`.
- [x] Passed `closeEditor()` to the `Alert` component.

1. `EditorNotification.js`

```js
  const [isClosed, setIsClosed] = React.useState(false)

  if (!hasEditingAccess(user, project) || isClosed) {
    return (<></>);
  }
  
  function closeEditor() {
    setIsClosed(true)
  }

  /*
  ...
  */

        <Alert handleClose={closeEditor} signal="notify">
```

- [x] Expected `handleClose` as a **void function** in `AlertProps` file.

2. `index.ts`
```ts
export interface AlertProps {
  signal?: 'notify' | 'success' | 'error';
  handleClose?: () => void;
}
```

- [x] In **Alert component**, involved `Icons.Close` in a **button** to select it in `StyledAlert.ts`

3. `Alert.tsx`
```tsx
  <button>
    <Icons.Close onClick={handleClose} />
   </button>
```

- [x] Removed button border and background and added cursor pointer behavior.

4. `StyledAlert.ts`
```ts
> button {
    border: none;
    background: none;
    cursor: pointer;
  }
```

### DIFF
--- a/packages/UI/src/components/molecules/Alert/Alert.tsx
+++ b/packages/UI/src/components/molecules/Alert/Alert.tsx
@@ -6,6 +6,7 @@ import type { AlertProps } from '.';
 
 const Alert: React.FC<React.PropsWithChildren<AlertProps>> = ({
   signal = 'notify',
+  handleClose,
   children,
   ...rest
 }) => {
@@ -21,7 +22,9 @@ const Alert: React.FC<React.PropsWithChildren<AlertProps>> = ({
         <></>
       )}
       <Typography css={{ flex: 1 }}>{children}</Typography>
-      <Icons.Close />
+      <button>
+        <Icons.Close onClick={handleClose} />
+      </button>
     </Container>
   );
 };
--- a/packages/UI/src/components/molecules/Alert/Alert.tsx
+++ b/packages/UI/src/components/molecules/Alert/Alert.tsx
@@ -4,14 +4,32 @@ import Typography from '../../atoms/Typography';
 import { Container } from './StyledAlert';
 import type { AlertProps } from '.';
 
+const variants = {
+  open: { opacity: 1, y: 0 },
+  closed: { opacity: 0, y: '-100%' },
+};
+
 const Alert: React.FC<React.PropsWithChildren<AlertProps>> = ({
   signal = 'notify',
   handleClose,
+  isInitallyOpen = true,
   children,
   ...rest
 }) => {
+  const [isOpen, setIsOpen] = React.useState(isInitallyOpen);
   return (
-    <Container signal={signal} {...rest}>
+    <Container
+      signal={signal}
+      {...rest}
+      variants={variants}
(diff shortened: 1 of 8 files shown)


packages/UI/src/components/molecules/Alert/Alert.tsx
```diff
@@ -21,7 +22,9 @@ const Alert: React.FC<React.PropsWithChildren<AlertProps>> = ({
         <></>
       )}
       <Typography css={{ flex: 1 }}>{children}</Typography>
-      <Icons.Close />
+      <button>
+        <Icons.Close onClick={handleClose} />
```
Enjoy2Live, line 26: @helioLJ this is a great addition! thanks for adding handleClose prop.
can you please make the onClick on the button rather than the Icon

packages/UI/src/components/molecules/Alert/StyledAlert.ts
```diff
@@ -28,4 +28,10 @@ export const Container = styled.div<Styles>`
       `;
     }}
   }
+
+  > button {
```
Enjoy2Live, line 32: it's recommended to write it this way: `& > button {`

packages/UI/src/components/molecules/Alert/index.ts
```diff
@@ -1,4 +1,5 @@
 export interface AlertProps {
   signal?: 'notify' | 'success' | 'error';
+  handleClose?: () => void;
```
Enjoy2Live, line 3: the type here is not so accurate `handleClose?: (e?: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void;` is more explicit

packages/UI/src/components/molecules/Alert/Alert.tsx
helioLJ: Sure! I it will be more concise this way, ty

packages/UI/src/components/molecules/Alert/StyledAlert.ts
helioLJ: Ok

packages/UI/src/components/molecules/Alert/index.ts
helioLJ: Thanks for your feedback, Mohammed! I'll make these changes soon as possible

Enjoy2Live: @helioLJ I made a PR to fix those and also add animation to the alert https://github.com/dev-launchers/dev-launchers-platform/pull/1253

PR #1277, opened 2023-07-19
Development/dev recruiters

ktesfaye:


packages/UI/src/components/atoms/Typography/Styled.Typography.ts
```diff
@@ -114,7 +114,7 @@ const label = `
     font-size: 0.875rem;
     font-weight: 400;
     line-height: 1.25rem;
-    text-transform: uppercase;
+    text-transform: none;
```
Enjoy2Live, line 117: @ktesfaye 
why are we changing text to be non uppercase globally here?

Enjoy2Live: please don't merge this until we ask design team about this

Enjoy2Live (changes requested):

Issue #1148, opened 2023-05-26
[Image section]As a potential member, I would like some visual elements in the join page so that I am engaged and interested to know further about Dev launchers


Srividya1394:

Srividya1394:
- Create two placeholders for images in the join page as shown in the image below
- The images to be included will be provided by design team
- ![image.png](https://images.zenhubusercontent.com/6460df08ff04bb3b6e33b654/aa015185-8147-427a-acf4-31fb554df759)

Issue #1147, opened 2023-05-24
[Introduction section] As a potential member, I want the introduction section of the join page to be precise and short so that I get an idea of what to expect further


Srividya1394:
- Include the following content in the introduction section of the join page to clearly inform the user about the purpose of the page
"Want to level up your tech skills and build world-changing products? "
_"Join our collaborative, global team as we build the future of technology. 
Check out our open roles below and apply"
![image.png](https://images.zenhubusercontent.com/6460df08ff04bb3b6e33b654/21b59bef-1ee1-4308-961b-848b8f9f3f08)

PR #1280, opened 2023-07-19
Dev recurit: add introduction and image section

Jamilsonh:
#1147 and #1148


apps/dev-recruiters/src/components/modules/NewJoinPageComponent/ContainerImage/StyledContainerImage.ts
```diff
@@ -1,13 +1,10 @@
 import styled from 'styled-components';
 
 export const HeaderImage = styled.img`
-  margin: 5%;
-  display: inline;
 `;
```
ktesfaye, line 4: Does the HeaderImage need to still be here, since its styles are removed?

PR #1325, opened 2023-08-08
rina/new1075-import-SVG-files

rinagobo:
1. Imported image sources for the three role cards as SVG file 
2. Created "RoleCard" file in the "Images" > "Onboarding" file to store the three images imported.


judeamos (approved):
Looks good

Issue #1281, opened 2023-07-19
[US0.0] Create a resources page


Devanshusp:
**Problem**:
- A user should be able to see a resources page
- Currently no separate resources page exists

**Goal**:
- Create a page in the following directory: devlaunchers.org/projects/resources
- Only signed in users can view this page (need to check user context)
- If a user is not signed in, ask user to sign in (there is an example of this on https://devlaunchers.org/users/me page, if the user is not logged in they are shown the following)
- ![image.png](https://images.zenhubusercontent.com/64921182b8a2370ddbf0f9b2/0dba72e8-5e56-4fde-985a-991c22928c1a)

**Additional Information**:
- Link to [Figma](https://www.figma.com/file/hS4Cvs69za7LNkT9BYaDlo/Resource-Page?type=design&node-id=0%3A1&mode=design&t=4lW8pI9q9HyVF7qF-1)
- Link to [Storybook](https://storybook.devlaunchers.org/?path=/story/docs-iconography--page)

PR #1326, opened 2023-08-08
Yacine/create resources page #1281

yacineKahlerras:
- created resource page
- added other unfinished components
- fixed milestones task undefined error that might have caused a pr build error


apps/site-projects/src/components/modules/Resources/ProjectResources/TeamLinks/ResourceCard.js
```diff
@@ -0,0 +1,15 @@
+import Image from "next/image";
+import Button from "../../../../../../../../packages/UI/src/components/atoms/Button";
```
Enjoy2Live, line 2: please use imports this way.
import { Button } from "@devlaunchers/components/src/components/atoms";

PR #1328, opened 2023-08-11
created projects display and select; reformed userDataContext

helioLJ:
removed fetches from userDataContext, realized it's unnecessary;

retrieving projects with getStaticProps like the other pages;

getting teams from user (the function is named findUserRoles, i'll rename later) with userId and list of all projects;

displaying these teams and allowing select between them (and changing the cardSelected state dynamically, so the page content will change too), in case user isn't in any team, nothing is displayed;

### DIFF
--- a/apps/site-projects/src/components/modules/Resources/YourProjects/YourProjects.js
+++ b/apps/site-projects/src/components/modules/Resources/YourProjects/YourProjects.js
@@ -1,38 +1,18 @@
 import Card from "../../../common/Card";
 import { ProjectContainer } from "../../Projects/StyledProjects";
+import { InnerProjectContainer, YourProjectsContainer } from "./StyledYourProjects";
 
 function YourProjects({ userProjects, selectedCard, setSelectedCard }) {
     return (
-        <div>
+        <YourProjectsContainer>
               <h3>Your Projects</h3>
-              <ul
-                style={{
-                  paddingLeft: '0',
-                  display: 'flex',
-                  flexWrap: 'wrap',
-                  gap: '1.5rem',
-                }}
-              >
+              <ul>
                 {userProjects ? (
                   userProjects.map((project) => {
                     return (
                       <ProjectContainer key={project.id}>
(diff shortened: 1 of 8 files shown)


apps/site-projects/src/components/modules/Resources/index.js
```diff
@@ -1,9 +1,34 @@
...
+      userTeams.push(data);
+    }
+  }
+
+  return userTeams.length > 0 ? userTeams : null;
+}
+
+function Resources({ projects }) {
+  const { userData } = useUserDataContext();
+  const userProjects = findUserTeams(userData.id, projects)
```
yacineKahlerras, line 29: Hey Helio i read the code really nice work, only here **userProjects** is _null_ and throws an error because it uses @Enjoy2Live 's user context data and for some reason it doesn't return any teams for him, but using @pyxld-kris 's default user data it works and returns the teams array

apps/site-projects/src/components/modules/Resources/YourProjects/YourProjects.js
```diff
@@ -17,25 +13,45 @@ function YourProjects({projects}) {
...
-                      />
-                    </ProjectContainer>
-                  );
-                })}
+                {userProjects ? (
+                  userProjects.map((project) => {
+                    return (
+                      <ProjectContainer key={project.id}>
+                        <div style={{ position: 'relative', width: '100%', height: '100%', border: selectedCard.id == project.id ? '6px solid #3A7CA5' : 'none' }}>
+                        <button style={{
```
yacineKahlerras, line 21: also here it works nicely i can select the projects and the project title bellow changes along with the border of the card, the only thing i would add here is using **StyledComponents** instead of adding styles in the **style prop** to seperate the css from the code and make it lighter, but otherwise overall good work !

yacineKahlerras (changes requested):

apps/site-projects/src/components/modules/Resources/YourProjects/YourProjects.js
helioLJ: thanks for the review, the changes were already made

yacineKahlerras:
nicely done !

Issue #1341, opened 2023-08-17
[US-DEV] Importing theme from UI and updating pages to use theme values


yacineKahlerras:
**Problem**

the  UI pachages components use new theme colors and in order to use those components we have to also import the theme the theme from UI packages.

**Goal**

- import theme from UI packages
- update New Builds pages to use those values

PR #1342, opened 2023-08-17
Importing the UI theme and updating pages to use colors from that theme #1341

yacineKahlerras:
- imported theme from UI packages
- updated pages to use colors from that UI packages theme


Kamui-Hebertt (approved):
Good job!

PR #1352, opened 2023-08-18
create people display and others changes

helioLJ:
- changed userDataContext id to Kris
- created PeopleDisplayContainer
- wrapped ProjectResources and TeamResources in SectionResources, so the design can stay synced with Figma

### DIFF
--- a/apps/site-projects/src/components/modules/Resources/TeamResources/TeamResources.js
+++ b/apps/site-projects/src/components/modules/Resources/TeamResources/TeamResources.js
@@ -1,22 +1,8 @@
 import PeopleDisplay from "./PeopleDisplay";
 import { TeamResourcesContainer } from "./StyledTeamResources";
 
-function getMembersAndLeaders(team) {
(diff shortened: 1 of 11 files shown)


apps/site-projects/src/components/modules/Resources/TeamResources/TeamResources.js
```diff
@@ -0,0 +1,41 @@
...
+
+    team.members.forEach(element => {
+        allTeam.push(element)
+    });
+
+    return allTeam
+}
+
+function TeamResources({ team }) {
+    const allTeam = getMembersAndLeaders(team)
```
yacineKahlerras, line 19: if you're just merging two arrays you can just use the spread operator instead of calling the function : 
```const allTeam = [...team.leaders,...team.members];```

helioLJ: This is totally true, I've commited your nice suggestion and i'll merge rn

PR #1367, opened 2023-09-03
Hotfix cannot login

dbradham:
This change will handle the new data formats from strapi v4 in the user profile page

### DIFF
--- a/apps/website/src/components/modules/UserProfile/UserProfile.js
+++ b/apps/website/src/components/modules/UserProfile/UserProfile.js
@@ -63,8 +63,8 @@ export default function UserProfile({ otherUser }) {
           setOpportunities(tempOpportunities);
         }
       })
-      .catch(() => {
-        console.error("Could not fetch project data");
+      .catch((e) => {
+        console.log("Could not fetch project data");
       });
   };
   React.useEffect(() => {
@@ -110,8 +110,8 @@ export default function UserProfile({ otherUser }) {
           setInterests(data);
         }
       })
-      .catch(() => {
-        console.error("Could not fetch interest data");
+      .catch((e) => {
+        console.log("Could not fetch interest data");
       });
   };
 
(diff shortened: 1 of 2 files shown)


chungthuang (approved):

apps/website/src/components/modules/UserProfile/UserProfile.js
```diff
@@ -18,6 +18,7 @@ import { Misc, UserInfo, UserSection, Wrapper } from './StyledUserProfile';
 import UserInterests from './UserInterests';
 import { useRouter } from "next/router";
 // import DiscordSection from "./DiscordSection/DiscordSection";
+import { cleanDataList } from '@devlaunchers/ideaspace/src/utils/StrapiHelper';
```
Enjoy2Live, line 21: I remember this function, I think it has to go to the utility shared package, because apps shouldn't depend on each other

apps/website/src/components/modules/UserProfile/UserProfile.js
```diff
@@ -58,8 +63,8 @@ export default function UserProfile({ otherUser }) {
           setOpportunities(tempOpportunities);
         }
       })
-      .catch(() => {
-        console.error("Could not fetch project data");
```
Enjoy2Live, line 62: Why changing this to console.log even though it's more of console.error

Enjoy2Live (changes requested):
Looks good for the most part, I've added only two requested changes

apps/website/src/components/modules/UserProfile/UserProfile.js
dbradham: Does this require changing the function to type script? I'm not sure how to proceed with that.

Enjoy2Live: @dbradham no, only moving it to the utility folder and exporting it

Issue #1337, opened 2023-08-16
[US2.4] Create a Team Meetings Display and its contents


zainab-aamir:
**Problem:**
A user should have access to all links to team’s weekly meetings

**Goal:**

- Create separate icons/display for every meeting 
- Link the displays to an external google meet/zoom/video conferencing link so that a user can click that to attend any meeting.
- For a UX user, the links should be to “Weekly Team Meetings” and “Product Meetings”
- For a Dev user, the links should be to “Weekly team meetings” and “Dev Meetings”

See the screenshot below for reference:
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBa29QIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--4a38026f7baeb1f2c57765054a639b100dfbe0d7/image.png)

**Additional Information:**
Link to[ Figma](https://www.figma.com/file/hS4Cvs69za7LNkT9BYaDlo/Resource-Page?type=design&node-id=468-4991&mode=design&t=a3G2C3d9XJJXwysw-0)

PR #1370, opened 2023-09-04
adding team meetings component #1337

yacineKahlerras:


apps/site-projects/src/components/modules/Resources/TeamMeetings/StyledTeamMeetings.js
```diff
@@ -0,0 +1,26 @@
...
+    width: 100%;
+
+    @media (min-width:40rem){
+        padding: 4rem 3rem;
+    }
+
+    > legend {
+        font-family: "Abel",sans-serif;
+        font-size: 2.3rem;
+        font-weight: bold;
```
jkoontz2010, line 23: do we have theme variables for fonts?

yacineKahlerras: ah yeah totally missed that, i updated the other legends from the other components as well, they all use ``theme.fonts.headline`` now

jkoontz2010 (approved):

PR #1400, opened 2023-09-17
Development/user profile 1258

sky1122:
- change the name 'one page' to 'congratulation page'
- Wrap another layer to style CongratulationPage.js
- Change StyledCongratulationPage.js
- Blocker: cannot make the img to black


sky1122:
try to update the code before create PR

PR #1418, opened 2023-09-26
Lamodot/new button

lamodots:
Refactored he NewButton component, made changes to the naming, removed the argsType since Typescript can handle it.


Enjoy2Live:
let's rename the files to not have "new"

Issue #1348, opened 2023-08-17
[Sign-in to Save later or Apply] As a DevLauncher, we want to keep track of the user who are applying to provide best user experience and ensure there is no spamming


Srividya1394:
- In the project details card, if the user is not signed in , include "sign in to save later or apply button"
- First project details card display :-
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBa2dXIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--1b5fbb7b21c9f787d8918a040de4fe6758338b8f/image.png)

After signing in, project details card display:-
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBa3dXIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--0ce8849bebbfda5a8557a560efbee9cc7b8f4e04/image.png)

Issue #1154, opened 2023-05-26
As a potential member, I want to be able to save the role I am interested in for future reference or if I want to return back later to complete the process so that I do not lose track of my previous activity


Srividya1394:
- Include save for later button in project detail pop-up dialog box
- This button will prompt the user to google sign up dialog box if the user is not logged in (use existing google sign-on logic)
- Once user is logged in, clicking on this button should save the role to the list associated with the user account so that the user can retrieve it in the future when they login

PR #1421, opened 2023-09-27
create save for later button

Jamilsonh:
- Create save later button 
- #1154
- Create the logic of sign-in to save later or Apply
- #1348


packages/UI/src/components/organisms/Navigation/MobileNavigation.tsx
```diff
@@ -58,7 +59,7 @@ const MobileNavigation = ({
                   <img
                     width="36"
                     height="33"
-                    src={user.profilePictureUrl}
+                    src={ProfilePhoto}
```
Enjoy2Live, line 62: This should be reverted as we're going to always show the logged-in user here rather than a static image

Enjoy2Live (changes requested):
@Jamilsonh not sure why this PR is still up, it touches many critical and global files and we need to take everyone's input before we merge it

PR #1471, opened 2023-10-12
Adjusted the shade movement

rinagobo:
1. Modified the 'animate' and 'transition' attribute for <m.div /> in PageOne.js.
2.  Modified const PageOneContainer = styled.div` div{width: } `; in StyledPageOne.js.


Enjoy2Live (approved):
LGTM
one of the few animations we have on the website, Thanks for your contribution!

PR #1478, opened 2023-10-21
Rina/1116 page two animation

rinagobo:
1. Implement an animate function for the three blurbs in the second onboarding tour page.(PageTwo.js)
2. Implement a button disability function.
    - Create new arguments called "disabled", "delayEnable" in stepperConfig property.
    - Create state hook for "buttonConfig.next.disabled"
    - Create "enableNextButton" object as a button enabling function.
    - Create "updateConfigs" object to contain all the state updates.


judeamos (approved):
All looks good to me, great work 👍🏾

PR #1515, opened 2023-11-16
Lamodot/notification

lamodots:
Completed the notification component task and make changes as suggested by PD team.


packages/UI/src/components/atoms/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,58 @@
...
+      lg: ' rounded-lg',
+      full: ' rounded-full',
+    },
+  },
+});
+
+interface AvatarProps extends VariantProps<typeof AvatarStyles> {
+  src: string;
+  alt: string;
+  delayMs: number;
```
Enjoy2Live, line 26: maybe we should have this as optional prop

PR #1621, opened 2024-01-14
Components/1605 checkbox

l1keafox:
Checkbox


packages/UI/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,45 @@
...
+      checked: {
+        false: 'hidden',
+      },
+    },
+  });
+
+  return (
+    <div className={checkbox({ className, state, checked })} {...props}>
+      <Checkmark className={checkmark({ checked })} />
+    </div>
```
Enjoy2Live, line 43: It would be better to use `<input>` checkbox [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input), even better is using radix ui but let's start simple and only do `<input>`

packages/UI/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,45 @@
...
+import Checkmark from './../../assets/icons/Checkmark';
+import type { CheckboxProps } from '.';
+
+export default function Checkbox({
+  checked = false,
+  className,
+  state,
+  ...props
+}: CheckboxProps) {
+  const checkbox = tv({
```
Enjoy2Live, line 11: tailwind variants (tv) declaration should always be outside the jsx function.
having it inside means it'll re-declare whenever the component re-renders which is going to reduce the performance.

packages/UI/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,45 @@
...
+        true: `${
+          state !== 'disable'
+            ? 'bg-brand-alt-nebula-500'
+            : 'border-grayscale-400 bg-grayscale-400'
+        }`,
+      },
+    },
+  });
+
+  const checkmark = tv({
```
Enjoy2Live, line 31: tailwind variants (tv) declaration should always be outside the jsx function.
having it inside means it'll re-declare whenever the component re-renders which is going to reduce the performance.

packages/UI/src/components/Checkbox/index.ts
```diff
@@ -0,0 +1,6 @@
+export { default as Checkbox } from './Checkbox';
+
+export interface CheckboxProps extends React.HTMLAttributes<HTMLDivElement> {
```
Enjoy2Live, line 3: nice attention to details with extending the needed props, big thumbs up!

packages/UI/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const checkbox = tv({
+    base: 'flex h-6 w-6 items-center justify-center rounded-sm border-2 border-solid border-brand-alt-nebula-500 bg-grayscale-50',
+    variants: {
+      state: {
+        focus:
+          'outline outline-2 outline-offset-2 outline-brand-alt-nebula-500',
+        hover:
+          'outline outline-4 outline-offset-0 outline-brand-alt-nebula-100',
+        disable: 'border-grayscale-500 bg-grayscale-200',
+      },
```
Enjoy2Live, line 20: These states can be handled better using [tailwind classes](https://tailwindcss.com/docs/hover-focus-and-other-states),
the current approach means the end user won't be able to toggle the hover state nor the focus state using their mouse/tab key for example.

Enjoy2Live (changes requested):
Very good start, I think the needed changes are minor but still needed so please go over my comments and if you questions I'll answer them here.
Thanks as always!

l1keafox:
@Enjoy2Live I'm not certain if I installed the radix-ui stuff correctly with yarn. Please look at the yarn files and let me know.

package.json
```diff
@@ -54,6 +54,8 @@
     "commit": "git-cz"
   },
   "dependencies": {
+    "@devlaunchers/components": "workspace:*",
+    "@radix-ui/react-checkbox": "1.0.4",
```
Enjoy2Live, line 58: you installed these `"@devlaunchers/components"` and `"@radix-ui/react-checkbox"` globally to whole monorepo. the root `package.json` should only include deps that will manage the whole experience of the monorepo, for example `rimraf` is installed globally because sometimes someone might want to delete every `node_modules` folder in the monorepo with the script `yarn nuke:node_modules` that can be found in the `package.json` in the root of the monorepo.

to fix this we need to remove `"@devlaunchers/components"` and `"@radix-ui/react-checkbox"` from the global `package.json` then run `yarn`

packages/UI/src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+};
+
+export const Disabled: Story = {
+  render: () => (
+    <div className="flex gap-3">
+      <Checkbox disabled />
+      <Checkbox disabled checked />
+    </div>
+  ),
+};
```
Enjoy2Live, line 48: I think we should create stories for each single state found in figma, for example:
`export const DefaultUnchecked = ....`
`export const DefaultChecked = ....`
`export const HoverUnchecked = ....`
etc

also adding [storybook-addon-pseudo-states](https://storybook.js.org/addons/storybook-addon-pseudo-states) addon to these will make it perfectly aligned to the design!

Enjoy2Live (changes requested):
GOOD JOB! the component itself renders just like Figma only thing left is making sure:
- [x] The stories are created properly
- [ ] The Checkmark gets re-exported to `packages/UI/src/assets/icons/index.ts`
- [ ] Configure tailwind.config.js to accept `data-state` rather than arbitrarily using `data-[state="checked"]`
- [ ] adding storybook-addon-pseudo-states to the stories that need them, example: `HoverChecked`.

I added more details in each chunk so if it doesn't make sense you can leave a comment on my requested changes. Thanks a ton Raymond :rocket:

packages/UI/src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+  component: Checkbox,
+  argTypes: {
+    checked: {
+      control: 'boolean',
+      description: 'Indicates whether the checkbox is checked',
+    },
+    disabled: {
+      control: 'boolean',
+      description: 'Indicates whether the checkbox is disabled',
+    },
```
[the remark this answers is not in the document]
l1keafox: @Enjoy2Live 
This adds in description in the middle column. 

I see in your Tabs.tsx you have it as a comment in your code. I'll change it too that format.

image

packages/UI/src/components/Checkbox/Checkbox.tsx
```diff
@@ -5,7 +5,7 @@ import CheckmarkImg from './../../assets/icons/Checkmark';
 const checkboxStyles = tv({
   slots: {
     rootStyle:
-      'flex h-6 w-6 items-center justify-center rounded-sm border-2 border-solid border-brand-alt-nebula-500 bg-grayscale-50 hover:outline hover:outline-4 hover:outline-offset-0 hover:outline-brand-alt-nebula-100 focus:outline focus:outline-2 focus:outline-offset-2 focus:outline-brand-alt-nebula-500 disabled:border-grayscale-500 disabled:bg-grayscale-200 disabled:hover:outline-none disabled:hover:outline-0 data-checked:bg-brand-alt-nebula-500 data-checked:disabled:bg-grayscale-200 hover:disabled:outline-0',
+      'flex h-6 w-6 items-center justify-center rounded-sm border-2 border-solid border-brand-alt-nebula-500 bg-grayscale-50 hover:outline hover:outline-4 hover:outline-offset-0 hover:outline-brand-alt-nebula-100 focus:outline focus:outline-2 focus:outline-offset-2 focus:outline-brand-alt-nebula-500 disabled:border-grayscale-500 disabled:bg-grayscale-200 disabled:hover:outline-none disabled:hover:outline-0 hover:disabled:outline-0 data-checked:bg-brand-alt-nebula-500 data-checked:disabled:bg-grayscale-200',
```
Enjoy2Live, line 8: why do we need `hover:disabled:outline-0` and `disabled:hover:outline-0` at the same time?

packages/UI/src/components/Checkbox/Checkbox.tsx
```diff
@@ -58,7 +58,9 @@ const checkboxStyles = tv({
  */
 
 export default function Checkbox({ className, ...props }: CheckboxProps) {
-  const { rootStyle, indicatorStyle } = checkboxStyles();
+  const { rootStyle, indicatorStyle } = checkboxStyles({
+    className,
```
Enjoy2Live, line 57: when someone passes a className prop, which element/slot would be affected here?

l1keafox: Oh both, fixing it now.

Enjoy2Live (approved):
Here's the long awaited approval!

Much appreciated, let's see what @dbradham @JulieMass @Nouri-Anouar @annaisaacks think about this

https://storybook.devlaunchers.org/components/1605Checkbox

PR #1629, opened 2024-01-17
1605Checkbox/suggestions

Enjoy2Live:
components/1605Checkbox


Enjoy2Live:
@l1keafox I'm going to close this as it's not going to be merged anyways.

packages/UI/src/components/Checkbox/Checkbox.tsx
```diff
@@ -5,49 +5,73 @@ import CheckmarkImg from './../../assets/icons/Checkmark';
...
+    required?: RadixCheckbox.CheckboxProps['required'];
+    /**
+     * The name of the checkbox. Submitted with its owning form as part of a name/value pair.
+     */
+    name?: RadixCheckbox.CheckboxProps['name'];
+    /**
+     * The value given as data when submitted with a name.
+     */
+    value?: RadixCheckbox.CheckboxProps['value'];
+  };
```
Enjoy2Live, line 54: I added every type from [radix-ui](https://www.radix-ui.com/primitives/docs/components/checkbox#root).
I had an issue with displaying the correct type on storybook docs for `checked`,`defaultChecked` props so I added `Type['source'] & NonNullable<unkown>` to them which seems to have fixed it.

try it out for yourself, remove `& NonNullable<unkown>` from the type `CheckboxProps` and observe the according prop on storybook docs page.

I think we should have this moved to your branch.

packages/UI/src/components/Checkbox/Checkbox.tsx
```diff
@@ -5,49 +5,73 @@ import CheckmarkImg from './../../assets/icons/Checkmark';
...
 
 /**
  * @description Created with the Checkbox component from Radix UI.
  * https://www.radix-ui.com/primitives/docs/components/checkbox
  */
 
-export default function Checkbox({ className, ...props }: CheckboxProps) {
-  const { rootStyle, indicatorStyle } = checkboxStyles({
-    checked: props.checked,
```
Enjoy2Live, line 44: I removed `checked: props.checked` because we already have `data-checked:` taking its role.
using tailwind variants for this is an over kill as tailwind itself can handle this by targeting the html attributes that are being mutated by radix ui.

packages/UI/src/components/Checkbox/Checkbox.tsx
```diff
@@ -5,49 +5,73 @@ import CheckmarkImg from './../../assets/icons/Checkmark';
 const checkboxStyles = tv({
   slots: {
     rootStyle:
-      'flex h-6 w-6 items-center justify-center rounded-sm border-2 border-solid border-brand-alt-nebula-500 bg-grayscale-50 hover:outline hover:outline-4 hover:outline-offset-0 hover:outline-brand-alt-nebula-100 focus:outline focus:outline-2 focus:outline-offset-2 focus:outline-brand-alt-nebula-500 disabled:border-grayscale-500 disabled:bg-grayscale-200 disabled:hover:outline-none disabled:hover:outline-0 hover:disabled:outline-0 data-checked:bg-brand-alt-nebula-500 data-checked:disabled:bg-grayscale-200',
+      'flex h-6 w-6 items-center justify-center rounded-sm border-2 border-solid border-brand-alt-nebula-500 bg-grayscale-50 focus-within:outline focus-within:outline-2 focus-within:outline-offset-2 focus-within:outline-brand-alt-nebula-500 hover:outline hover:outline-4 hover:outline-offset-0 hover:outline-brand-alt-nebula-100 disabled:border-grayscale-500 disabled:bg-grayscale-200 disabled:hover:outline-none disabled:hover:outline-0 hover:disabled:outline-0 data-checked:bg-brand-alt-nebula-500 data-checked:disabled:bg-grayscale-200',
```
Enjoy2Live, line 8: The only thing I changed here is `focus:` to `focus-within:`, I actually made a mistake here as I should've changed it to `focus-visible:`.

[Read this](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#examples) which demonstrates the difference between `focus:` and `focus-visible:` and let me know what you think.

packages/UI/src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -18,18 +21,19 @@ type Story = StoryObj<typeof Checkbox>;
...
+export const DefaultChecked: Story = {
+  render: (args) => <Checkbox {...args} defaultChecked />,
+};
+
+export const DisabledUnchecked: Story = {
+  render: (args) => <Checkbox {...args} disabled />,
+};
+
+export const DisabledChecked: Story = {
+  render: (args) => <Checkbox {...args} disabled defaultChecked />,
```
Enjoy2Live, line 38: since the component changes its behaviors **using props** only in four cases, I think we should have all the `export`ed stories moved to your branch as shown.

packages/UI/src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -4,9 +4,12 @@ import Checkbox from './Checkbox';
...
   component: Checkbox,
-  args: {
-    checked: false,
-    disabled: false,
+  argTypes: {
+    checked: { control: 'radio', options: [true, false, 'indeterminate'] },
+    defaultChecked: {
+      control: 'radio',
+      options: [true, false, 'indeterminate'],
+    },
```
Enjoy2Live, line 12: I added those to have better control on storybook's control tab when changing the props.

you can find the docs for changing the controls however you see fit.
https://storybook.js.org/docs/essentials/controls#annotation

packages/UI/src/components/Checkbox/Checkbox.mdx
```diff
@@ -10,6 +10,48 @@ Created with the Checkbox component from Radix UI.
...
+    <iframe
+      height="250"
+      width="400"
+      src="https://www.figma.com/embed?embed_host=devlaunchers.org&url=https://www.figma.com/file/EwzuhhvTulvFRMvhTD5VAh/DL-Universal-Design-System?type=design&node-id=9091-8801&mode=design&t=KOqNk3eFsNeE9tUj-4"
+      allowFullScreen
+    />
+  </div>
+</div>
+<Controls />
+<Stories />
```
Enjoy2Live, line 57: I think we should have all mdx file moved over

packages/UI/src/components/Checkbox/Checkbox.mdx
```diff
@@ -1,6 +1,6 @@
-import { Canvas, Meta, Controls } from '@storybook/blocks';
-
-import * as CheckboxStories from './Checkbox.stories';
+import { Canvas, Meta, Controls, Stories } from '@storybook/blocks';
+import * as CheckboxStories from './Checkbox.stories.tsx';
+import Checkbox, { checkboxStyles } from './Checkbox.tsx';
```
Enjoy2Live, line 3: it's preferred to keep the imports without the file extension. 
in this case it's better to have 
`import * as CheckboxStories from './Checkbox.stories';`
instead of 
`import * as CheckboxStories from './Checkbox.stories.tsx';`
as an example.

packages/UI/package.json
```diff
@@ -20,6 +20,7 @@
     "react-burger-menu": "^3.0.8",
     "react-dom": "^17.0.2",
     "react-scripts": "5.0.1",
+    "storybook-addon-designs": "6.3.1",
```
Enjoy2Live, line 23: This plugin allows you to render a component coming from figma, linked to a specific story.

for example you can link `DefaultChecked` story to the same exact state from Figma and then designers would be able to compare the design to the implementation right within storybook.
here's the [docs](https://storybook.js.org/addons/@storybook/addon-designs/)

packages/UI/.storybook/preview.jsx
```diff
@@ -66,7 +66,7 @@ export const decorators = [
   ),
 ];
 export const parameters = {
-  actions: { argTypesRegex: "^on[A-Z].*" },
+  actions: { argTypesRegex: '^on.*' },
```
Enjoy2Live, line 69: This change will make storybook observe all the events that start with `on` and display/console log them on the actions tab when the story viewer fires that event by interacting with the component.

pretty helpful to know what events a component supports with this
Example: someone wants to know what is the event responsible for changing the component state from checked to unchecked, they go and interact with the component to trigger that behavior then get to know the event that was fired from the actions tab.

packages/UI/.storybook/main.ts
```diff
@@ -33,9 +33,9 @@ const config: StorybookConfig = {
         esModuleInterop: false,
       },
       shouldExtractLiteralValuesFromEnum: true,
-      propFilter: () => true,
+      propFilter: (prop) => (prop.parent ? !/node_modules/.test(prop.parent.fileName) : true),
```
Enjoy2Live, line 36: This will filter out props that are coming `node_modules` from being in the docs.

packages/UI/.storybook/main.ts
```diff
@@ -33,9 +33,9 @@ const config: StorybookConfig = {
...
     },
-    reactDocgen: 'react-docgen',
+    reactDocgen: 'react-docgen-typescript',
```
Enjoy2Live, line 38: `react-docgen-typescript` is more advanced so I think we should keep `react-docgen-typescript` in

packages/UI/src/components/Checkbox/Checkbox.tsx
```diff
@@ -5,49 +5,73 @@ import CheckmarkImg from './../../assets/icons/Checkmark';
...
-  variants: {
-    checked: {
-      true: {
-        rootStyle: 'bg-brand-alt-nebula-500',
-      },
-      false: {
-        indicatorStyle: 'hidden',
-      },
-    },
-  },
```
Enjoy2Live, line 20: tailwind is already doing this with `data-checked:` modifier so I removed it from tailwind-variants

Enjoy2Live (commented):
@l1keafox , feel free discuss/ask questions about anything I made here

PR #1636, opened 2024-01-23
components/pagination

Enjoy2Live:
development/components


packages/UI/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,111 @@
...
+const PaginationItem = React.forwardRef<
+  HTMLLIElement,
+  React.ComponentProps<'li'>
+>(({ className, ...props }, ref) => (
+  <li ref={ref} className={cn('', className)} {...props} />
+));
+PaginationItem.displayName = 'PaginationItem';
+
+const paginationLinkStyles = tv({
+  base: 'inline-flex flex-col items-start gap-2,5 rounded-lg px-4 py-2 font-nunito-sans text-base font-normal leading-6',
```
l1keafox, line 45: gap-2,5

packages/UI/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,111 @@
...
+  ChevronLeft,
+} from 'lucide-react';
+import Link from 'next/link';
+import * as React from 'react';
+
+import type { VariantProps } from 'tailwind-variants';
+import { tv } from 'tailwind-variants';
+import { cn } from '../../utils/classesMerger';
+
+const Pagination = ({ className, ...props }: React.ComponentProps<'nav'>) => (
```
l1keafox, line 14: Is there any standard on using function over using a const variable for a component?  I see forwardRef needs to be this pattern

packages/UI/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,111 @@
...
+  ChevronsRight,
+  ChevronRight,
+  ChevronLeft,
+} from 'lucide-react';
+import Link from 'next/link';
+import * as React from 'react';
+
+import type { VariantProps } from 'tailwind-variants';
+import { tv } from 'tailwind-variants';
+import { cn } from '../../utils/classesMerger';
```
l1keafox, line 12: What is the benefit from having twMerge in this utils file? Seems like just an extra file/import just to implement twMerge as a different file.

packages/UI/src/components/Pagination/Pagination.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+ *👇 Render functions are a framework specific feature to allow you control on how the component renders.
+ * See https://storybook.js.org/docs/react/api/csf
+ * to learn how to use render functions.
+ */
+
+export const Primary: Story = {
+  render: () => (
+    <Pagination>
+      <PaginationContent>
+        <div className="flex gap-2">
```
l1keafox, line 28: We can remove all the divs and move the className to PaginationContent

packages/UI/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,111 @@
...
+
```
l1keafox, line 13: Add Jsdocs for pagination, also instructions that using href URL params is necessary for this pagination.

l1keafox (commented):
Just some comments and ideas.

packages/UI/src/components/Pagination/Pagination.tsx
Enjoy2Live: that's the alternative we use for `gap-2.5` coming from tailwind.
it's the value of gap between `gap-2` and `gap-3`.
I wish we were able to have with a dot rather than comma but Tokens-studio doesn't accept dots in the name of the toke

packages/UI/src/components/Pagination/Pagination.tsx
Enjoy2Live: yeah I prefer using const personally but we don't restrict devs on using it, maybe I should make eslint warn users for not using `const` way

packages/UI/src/components/Pagination/Pagination.tsx
Enjoy2Live: `cn` does the same thing twMerge but does a bit more by enabling objects to be passed over.
for example let's say some is using the `<Pagination />` and wants a specific style applied only on a specific condition: `<Pagination className={{'bg-black': someCondition}}>...</Pagination>`.
this is only possible using `cn` but if we use twMerge they can't pass object with condition.

packages/UI/src/components/Pagination/Pagination.tsx
Enjoy2Live: good call, thanks!

packages/UI/src/components/Pagination/Pagination.stories.tsx
Enjoy2Live: using something like grid? not sure what you mean here.
can you please give me an example, I don't like these divs there either!

l1keafox: You can just put the className in the Pagination Content and get rid of all the divs. The divs don't really do anything in this situation, you are splitting them, but nothing special with layout with the split divs.

Enjoy2Live: the split divs actually change the layout as they group the arrows left, numbers, arrows right.
each div acts as a flex container that puts a gap between each element by `0.5rem` aka `gap-2`.

`<PaginationContent />` is already a flex container with gap of `2rem` aka `gap-8` so I'm not sure how I can do the same without those extra divs, here's a screenshot explaining the current implementation!
image

l1keafox: Oh sorry, yeah, i didn't think that gap-8 created space between the three divs.  Ignore me then!

Enjoy2Live:
@l1keafox I updated my pagination PR to have jsdocs. thanks for the review as always

PR #1637, opened 2024-01-25
Components/931 modal

### LINKED ISSUE #931: implement Modal Container/Shower hook
https://www.figma.com/file/EwzuhhvTulvFRMvhTD5VAh/DL-Universal-Design-System?type=design&node-id=8375-7152&mode=design&t=2V2qv1p8IpuvDF96-4

l1keafox:
Resolves #931


packages/UI/src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,80 @@
...
+   */
+  header: React.ReactNode;
+  /**
+   * content: React.ReactNode - the element that will be rendered in the middle of the modal
+   */
+  content: React.ReactNode;
+  /**
+   * footer: React.ReactNode - the element that will be rendered at the bottom of the modal
+   */
+  footer: React.ReactNode;
```
Enjoy2Live, line 26: let's use the same description for these props from https://www.radix-ui.com/primitives/docs/components/dialog#root

packages/UI/src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,80 @@
...
+  trigger,
+  footer,
+  content,
+  className,
+  hasCloseBtn = true,
+}: ModalProps) => {
+  const { rootStyle, overlayStyle } = modalStyles();
+  return (
+    <Dialog.Root>
+      <Dialog.Trigger asChild>{trigger}</Dialog.Trigger>
```
Enjoy2Live, line 50: asChild is a bit scary as it could break the accessibility so it's a good idea to tell somewhere that `trigger` prop should a inherently a \<button />

packages/UI/src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,80 @@
...
+            <Dialog.Close asChild>
+              <button
+                className="absolute right-8 top-8 flex h-7 w-7 items-center justify-center rounded-md bg-white drop-shadow-xl"
+                aria-label="Close"
+              >
+                <X />
+              </button>
+            </Dialog.Close>
+          )}
+          <div className="my-6 min-h-36 w-full">{content}</div>
```
Enjoy2Live, line 65: eslint error: `Classname 'min-h-36' is not a Tailwind CSS class!`

packages/tailwind-constructor/tailwind.config.js
```diff
@@ -24,13 +24,19 @@ module.exports = withTV({
...
     fontFamily: filterTokensByType('fontFamilies', tokens),
     spacing: filterTokensByType('spacing', tokens),
     screens: {
       'sm': '640px',
       'md': '768px',
       'lg': '1024px',
+    },
+    extend: {
+      minWidth: {
+        128: "512px",
```
Enjoy2Live, line 38: why are we having this new value? we should stick to what the design team give us

packages/UI/src/components/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,217 @@
...
+  render: () => (
+    <Modal
+      trigger={PrimaryTriggerBtn}
+      header={PrimaryHeader}
+      content={PrimaryContent}
+      footer={PrimaryFooter}
+      hasCloseBtn={false}
+    />
+  ),
+};
```
Enjoy2Live, line 217: the code to shown to use these stories doesn't make alot of sense, maybe we should find a solution to make it show jsx rather than `PrimaryTriggerBtn` for example
image

Enjoy2Live (changes requested):
awesome start, we need to just find a solution for the docs not showing jsx but rather variables like `trigger={PrimaryTriggerBtn}`.

also if those stories you created `Primary` and `PrimaryNoCloseBtn` have designs please include using 
```
parameters: {
    design: {
      type: "figma",
      url: "https://link.to.node.that.represent.this.story",
    },
  },
  ```

File: packages/UI/src/components/Modal/Modal.tsx (hunk 3 above)
Comment (l1keafox) [reply packages/UI/src/components/Modal/Modal.tsx:65]: @Enjoy2Live it exists in the docs. 
https://tailwindcss.com/docs/min-height

File: packages/tailwind-constructor/tailwind.config.js (hunk 4 above)
Comment (l1keafox) [reply packages/tailwind-constructor/tailwind.config.js:38]: Any suggestions then on creating a min-width for the 448px? I used 512px because it was double of min-w-64 (256px). and didn't want to use [].

File: packages/UI/src/components/Modal/Modal.tsx (hunk 3 above)
Comment (Enjoy2Live) [reply packages/UI/src/components/Modal/Modal.tsx:65]: @l1keafox indeed, I need to debug the eslint rule

File: packages/tailwind-constructor/tailwind.config.js (hunk 4 above)
Comment (Enjoy2Live) [reply packages/tailwind-constructor/tailwind.config.js:38]: how did you come up with these? `max-md:w-11/12 md:min-w-128`. mind telling me the process of extracting them from the

File: packages/UI/src/components/Modal/Modal.tsx (hunk 3 above)
Comment (Enjoy2Live) [reply packages/UI/src/components/Modal/Modal.tsx:65]: does eslint say the same thing on your

Comment (Enjoy2Live) [reply packages/UI/src/components/Modal/Modal.tsx:65]: fixed by updating the eslint plugin as well as tailwind itself. nevermind!

File: packages/tailwind-constructor/tailwind.config.js (hunk 4 above)
Comment (l1keafox) [reply packages/tailwind-constructor/tailwind.config.js:38]: at small screens I wanted the modal took up most of width of the screen, that's why I did 11/12 width. I realize I also have mx-2 or something too. 

md:min-w-128, was that I thought 128 or 512px was going to be a common use size, something we can extend tailwind for. w-64 exists, so I was just thinking double of that would be something we would want more in the future.  Just so the modal always have a minimum size, which I grabbed from figma.

Comment (l1keafox) [reply packages/tailwind-constructor/tailwind.config.js:38]: image

Comment (Enjoy2Live) [reply packages/tailwind-constructor/tailwind.config.js:38]: @Nouri-Anouar do you want to add this new value to tokens studio?

File: packages/UI/src/components/Modal/Modal.stories.tsx (hunk 5 above)
Comment (l1keafox) [reply packages/UI/src/components/Modal/Modal.stories.tsx:217]: image

File: packages/UI/src/components/Modal/Modal.stories.tsx [hunk 6]
```diff
@@ -0,0 +1,527 @@
...
+export default meta;
+type Story = StoryObj<typeof Modal>;
+
+/*
+ *👇 Render functions are a framework specific feature to allow you control on how the component renders.
+ * See https://storybook.js.org/docs/react/api/csf
+ * to learn how to use render functions.
+ */
+
+export const Primary: Story = {
```
Enjoy2Live, line 27: This includes all different modals in one story which doesn't really make sense anymore since they're now their separate stories, Storybook should always be a mirror to what we have in Figma!

I suggest this story obj getting deleted

packages/UI/src/components/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,527 @@
...
+  },
+  parameters: {
+    design: {
+      type: 'figma',
+      url: 'https://www.figma.com/file/EwzuhhvTulvFRMvhTD5VAh/DL-Universal-Design-System?node-id=8769%3A6382',
+    },
+  },
+};
+
+export const EmptyModal: Story = {
```
Enjoy2Live, line 221: This should be named Default (same as Figma)
image

packages/UI/src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,82 @@
...
+              </button>
+            </Dialog.Close>
+          )}
+          <div className="min-h-36 min-w-96 my-6 w-full">{content}</div>
+          <div className="w-full min-w-96">{footer}</div>
+        </Dialog.Content>
+      </Dialog.Portal>
+    </Dialog.Root>
+  );
+};
```
Enjoy2Live, line 72: can we decouple these like in shadcn and radix docs?
It can be confusing to the consumer reading the docs of radix then using this as radix has these separately exported where here you have to pass them as props

packages/UI/src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+export default Modal;
+
+const modalStyles = tv({
+  slots: {
+    rootStyle:
+      'min-h-80 fixed left-1/2 top-1/2 z-50 mx-auto flex -translate-x-1/2 -translate-y-1/2 flex-col items-center justify-between rounded-lg bg-white p-8 max-md:w-11/12',
+    overlayStyle: `fixed inset-0 bg-Light-Overlay`,
+  },
+});
```
Enjoy2Live, line 82: let's move this to the top of the file, I have no idea how yours worked considering js doesn't hoist anything but functions

packages/UI/src/components/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,527 @@
...
+        <Modal
+          trigger={EmptyTriggerBtn}
+          header={EmptyHeader}
+          content={EmptyContent}
+          footer={EmptyFooter}
+        />
+      </div>
+    );
+  },
+};
```
Enjoy2Live, line 253: the empty modal doesn't render like how it is in figma

packages/UI/src/components/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,527 @@
...
+
+export const Subscribe: Story = {
```
Enjoy2Live, line 255: in Figma we have this variant named "Newsletter" rather than "Subscribe"

packages/UI/src/components/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,527 @@
...
+          </Button>
+        </Close>
+        <Close>
+          <Button buttonType="primary" buttonSize="standard">
+            {' '}
+            Subscribe{' '}
+          </Button>
+        </Close>
+      </div>
+    );
```
Enjoy2Live, line 312: why these jsx elements are inside variables? it doesn't look readable in the code on storybook docs nor serves any purpose in this file.

packages/UI/src/components/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,527 @@
...
+          header={ToSHeader}
+          content={ToSContent}
+          footer={ToSFooter}
+        />
+      </div>
+    );
+  },
+};
+
+export const RoleModal: Story = {
```
Enjoy2Live, line 392: let's have this named "Roles" to match Figma better.

Enjoy2Live (changes requested):
The component looks the same as in Figma, I think it needs to be more modular like in radix and shadcn to accommodate any use case and be relevant to radix docs.

packages/UI/src/components/Modal/Modal.stories.tsx
l1keafox: Because it's more readable in storybook than having them directly in the props.

packages/UI/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -47,11 +47,14 @@ export const Default: Story = {
           <DialogContent>
             <DialogHeader>
               <DialogTitle>
-                <div className="h-8 w-96 bg-white uppercase" />
+                {/* eslint-disable-next-line tailwindcss/no-arbitrary-value */}
```
Enjoy2Live, line 50: eslint-disable-next-line tailwindcss/no-arbitrary-value should be avoided as much as possible, just remove it.

eslint-disable-next-line tailwindcss/no-arbitrary-value means I can't notice this viloation when I run `yarn workspace @devlaunchers/components lint` which ultimately means the design tokens won't get updated and other devs will keep using arbitrary-value in their apps.

The only way I know developers are struggling with finding a token is through tailwindcss/no-arbitrary-value so disabling it makes me blind to what's missing and subsequently makes Anouar and Design Team blind about this missing token!

packages/UI/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -47,11 +47,14 @@ export const Default: Story = {
...
+                <div className="h-8 w-[448px] bg-white uppercase" />
               </DialogTitle>
             </DialogHeader>
-            <div className="h-36 w-96 bg-grayscale-100"></div>
-            <div className="h-8 w-96 bg-white"></div>
+            {/* eslint-disable-next-line tailwindcss/no-arbitrary-value */}
```
Enjoy2Live, line 54: eslint-disable-next-line tailwindcss/no-arbitrary-value should be avoided as much as possible, just remove it.

eslint-disable-next-line tailwindcss/no-arbitrary-value means I can't notice this viloation when I run `yarn workspace @devlaunchers/components lint` which ultimately means the design tokens won't get updated and other devs will keep using arbitrary-value in their apps.

The only way I know developers are struggling with finding a token is through tailwindcss/no-arbitrary-value so disabling it makes me blind to what's missing and subsequently makes Anouar and Design Team blind about this missing token!

packages/UI/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -47,11 +47,14 @@ export const Default: Story = {
...
+            <div className="h-36 w-[448px] bg-grayscale-100"></div>
+            {/* eslint-disable-next-line tailwindcss/no-arbitrary-value */}
```
Enjoy2Live, line 56: eslint-disable-next-line tailwindcss/no-arbitrary-value should be avoided as much as possible, just remove it.

eslint-disable-next-line tailwindcss/no-arbitrary-value means I can't notice this viloation when I run `yarn workspace @devlaunchers/components lint` which ultimately means the design tokens won't get updated and other devs will keep using arbitrary-value in their apps.

The only way I know developers are struggling with finding a token is through tailwindcss/no-arbitrary-value so disabling it makes me blind to what's missing and subsequently makes Anouar and Design Team blind about this missing token!

packages/UI/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -78,11 +81,14 @@ export const DefaultWithNoClose: Story = {
           <DialogContent hasCloseBtn={false}>
             <DialogHeader>
               <DialogTitle>
-                <div className="h-8 w-96 bg-white uppercase" />
+                {/* eslint-disable-next-line tailwindcss/no-arbitrary-value */}
```
Enjoy2Live, line 84: eslint-disable-next-line tailwindcss/no-arbitrary-value should be avoided as much as possible, just remove it.

eslint-disable-next-line tailwindcss/no-arbitrary-value means I can't notice this viloation when I run `yarn workspace @devlaunchers/components lint` which ultimately means the design tokens won't get updated and other devs will keep using arbitrary-value in their apps.

The only way I know developers are struggling with finding a token is through tailwindcss/no-arbitrary-value so disabling it makes me blind to what's missing and subsequently makes Anouar and Design Team blind about this missing token!

packages/UI/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -78,11 +81,14 @@ export const DefaultWithNoClose: Story = {
...
+                <div className="h-8 w-[448px] bg-white uppercase" />
               </DialogTitle>
             </DialogHeader>
-            <div className="h-36 w-96 bg-grayscale-100"></div>
-            <div className="h-8 w-96 bg-white"></div>
+            {/* eslint-disable-next-line tailwindcss/no-arbitrary-value */}
```
Enjoy2Live, line 88: eslint-disable-next-line tailwindcss/no-arbitrary-value should be avoided as much as possible, just remove it.

eslint-disable-next-line tailwindcss/no-arbitrary-value means I can't notice this viloation when I run `yarn workspace @devlaunchers/components lint` which ultimately means the design tokens won't get updated and other devs will keep using arbitrary-value in their apps.

The only way I know developers are struggling with finding a token is through tailwindcss/no-arbitrary-value so disabling it makes me blind to what's missing and subsequently makes Anouar and Design Team blind about this missing token!

packages/UI/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -78,11 +81,14 @@ export const DefaultWithNoClose: Story = {
...
+            <div className="h-36 w-[448px] bg-grayscale-100"></div>
+            {/* eslint-disable-next-line tailwindcss/no-arbitrary-value */}
```
Enjoy2Live, line 90: eslint-disable-next-line tailwindcss/no-arbitrary-value should be avoided as much as possible, just remove it.

eslint-disable-next-line tailwindcss/no-arbitrary-value means I can't notice this viloation when I run `yarn workspace @devlaunchers/components lint` which ultimately means the design tokens won't get updated and other devs will keep using arbitrary-value in their apps.

The only way I know developers are struggling with finding a token is through tailwindcss/no-arbitrary-value so disabling it makes me blind to what's missing and subsequently makes Anouar and Design Team blind about this missing token!

packages/UI/src/components/Dialog/Dialog.mdx
```diff
@@ -19,164 +19,162 @@ import {
...
+  </Dialog>
+  <Dialog>
+    <DialogTrigger>
+      <Button buttonType="primary" buttonSize="standard">
+        Subscribe!
+      </Button>
+    </DialogTrigger>
+    <DialogContent >
+      <DialogHeader>
+        {/* eslint-disable-next-line tailwindcss/no-arbitrary-value */}
```
Enjoy2Live, line 90: eslint-disable-next-line tailwindcss/no-arbitrary-value should be avoided as much as possible, just remove it.

eslint-disable-next-line tailwindcss/no-arbitrary-value means I can't notice this viloation when I run `yarn workspace @devlaunchers/components lint` which ultimately means the design tokens won't get updated and other devs will keep using arbitrary-value in their apps.

The only way I know developers are struggling with finding a token is through tailwindcss/no-arbitrary-value so disabling it makes me blind to what's missing and subsequently makes Anouar and Design Team blind about this missing token!

Enjoy2Live (changes requested):

Enjoy2Live:
Resolves #931

Issue #840, opened 2023-02-11
[US - DEV] As AMELIA HARRIS, I want to see an “idea owner” tag in the comment feed so that I can visually identify Miles's comments.


JulieMass:
### Considerations:
Miles’s comments will be highlighted in the comment feed with an “idea owner” tag to differentiate these from the other community members. 

Note: in the future we’ll be having “editor” tags for people that have edit permission on IdeaSpace’s content.

Please refer to UX designs in  [Figma](https://www.figma.com/file/hEDIHsBZYa6hmK8As5mcjy/%5BDL---IdeaSpace%5D-UX-designs?node-id=2645%3A5947&t=5fHSxtzyVh1IzGWL-1)

### Acceptance criteria:
- An “idea owner” tag is added next to all idea owners username in the comment feed
- Layout is web responsive and it works well for mobile, tablets and desktop breakpoints

----------------------------------------------------
### Sub-tasks:
- [X] UX design
- [ ] Development
- [ ] QA

----------------------------------------------------
![image.png](https://images.zenhubusercontent.com/63400f3d1e79614e45d9ebd9/68c2275a-a117-4b0e-ba79-a1fa33c739a7)

JulieMass:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/IdeaSpace-Team-6350185d9843157427cc900b/issues/dev-launchers/dev-launchers-platform/840?planning-poker) with Zenhub @alicefate @dbradham @jordanmcousins @JulieMass @Nouri-Anouar @RETprojects

dbradham:
@alicefate do you know if we currently have these fields available in the backend for an idea?

Nouri-Anouar:
@RETprojects 
I created an ideaspace role component [here](https://www.figma.com/file/hEDIHsBZYa6hmK8As5mcjy/%5BDL---IdeaSpace%5D-UX-designs?node-id=4119-10172&t=YhgxtYBkawRoIBiW-4)

I used colors from the styleguide that you can also find in the global theme. You can see which ones I used if you select an element and go to inspect tab
![image](https://user-images.githubusercontent.com/79311912/226862804-4eb04cc0-d75d-4d7a-b372-fa1d91ed3576.png)

let me know if you have questions.

@JulieMass  I've only designed two variants, orange and blue, do you think that is enough for now? considering tags that we discussed earlier.

JulieMass:
[...]

PR #1642, opened 2024-01-28
New idea owner label

RETprojects:
In support of ticket #840, I have added a dynamic idea owner label to SingleComment.


dbradham (approved):
Looks good!

Issue #913, opened 2023-03-15
[US - DEV] As AMELIA HARRIS, I want to follow an idea from the workshopping page so that I can easily get back to it from my saved list.


JulieMass:
### Considerations:
Ability to follow an idea from the [workshopping page](https://devlaunchers.org/ideaspace/browse).  This task is limited to backend development only, while the frontend will be addressed in a separate ticket.

Please refer to the Workshopping page UX designs in [Figma](https://www.figma.com/file/hEDIHsBZYa6hmK8As5mcjy/%5BDL---IdeaSpace%5D-UX-designs?node-id=2645%3A5947&t=8soOirvjw3YvrVlo-1) 

### IMPORTANT ⚠️:
We think #688 is a related ticket from the Product Enablement team (please verify with @Enjoy2Live) 

### Acceptance criteria:
-  Remove our (old) "Save" button from all our Ideas cards in the [Ideas page](https://devlaunchers.org/ideaspace/browse) and the [IdeaSpace's dashboard](https://devlaunchers.org/ideaspace/dashboard), and continue working on the backend for the 'Ideas I Follow' as the logic will be the same.

----------------------------------------------------
### **Sub-tasks:**
- [X] UX design
- [ ] Design Enhancement for the "vote", "Follow" and "Save" in the workshopping page (ticket #1586)
- [ ] Back-end Development
- [ ] Front-end Development 
- [ ] QA

----------------------------------------------------
<img width="968" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd2hpQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--1563d5bcd5d11f00c4232bd0f0f12d35077f41d2/image.png" alt="image.png" />

"Ideas I follow" list in the dashboard page:
<img width="494" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeHhpQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--5aaaea06ca3db2cddbd93c37f63bc1c661f1b6b9/image.png" alt="image.png" />

JulieMass:
[...]

PR #1643, opened 2024-01-28
Remove save idea from ideaspace

Rutuja177:
This removed the frontend code for save idea on the browse ideas page.

This PR is related to #913


dbradham (approved):
Looks good!

PR #1656, opened 2024-02-03
Components/avatar

lamodots:
This is the global Avatar component


packages/UI/src/components/atoms/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+const AvatarStyles = tv({
+  base: ' h-12 w-12 justify-center   rounded-full bg-white  bg-contain text-center align-middle text-base font-medium',
+});
+
+interface AvatarProps
+  extends VariantProps<typeof AvatarStyles>,
+    Avatar.AvatarProps {
+  src: string;
+  fallback: React.JSX.Element;
```
Enjoy2Live, line 13: we dont need fallback anymore

packages/UI/src/components/atoms/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,60 @@
...
+  alt: string;
+  delayMs?: Avatar.AvatarFallbackProps['delayMs'];
+  onLoadingStatusChange?: Avatar.AvatarImageProps['onLoadingStatusChange'];
```
Enjoy2Live, line 16: it would be nice to have the props with description based on what https://www.radix-ui.com/primitives/docs/components/avatar already describes them

packages/UI/src/components/atoms/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,60 @@
+import * as Avatar from '@radix-ui/react-avatar';
+
+import { tv, type VariantProps } from 'tailwind-variants';
+
+const AvatarStyles = tv({
+  base: ' h-12 w-12 justify-center   rounded-full bg-white  bg-contain text-center align-middle text-base font-medium',
+});
+
+interface AvatarProps
+  extends VariantProps<typeof AvatarStyles>,
```
Enjoy2Live, line 10: we don't need to extend VariantProps as this component doesn't have variants

packages/UI/src/components/atoms/Avatar/Avatar.stories.tsx
```diff
@@ -0,0 +1,39 @@
+import type { Meta, StoryObj } from '@storybook/react';
+import Avatar from './Avatar';
+
+const meta: Meta<typeof Avatar> = {
+  component: Avatar,
+  title: 'Atoms/Avatar',
```
Enjoy2Live, line 6: let's remove this title property

package.json
```diff
@@ -54,6 +54,7 @@
     "commit": "git-cz"
   },
   "dependencies": {
+    "@radix-ui/react-avatar": "1.0.4",
```
Enjoy2Live, line 57: move this to packages/UI/package.json and remove it from here. then run `yarn`

packages/UI/src/components/atoms/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,60 @@
...
+  onClick?: () => void;
+}
+
+const AvatarComponent = ({
+  src,
+  alt,
+  delayMs,
+
+  onLoadingStatusChange,
+  onClick,
```
Enjoy2Live, line 26: we don't have to explicitly use onClick, let's only have it coming from ...props

Enjoy2Live (changes requested):

PR #1664, opened 2024-02-09
components/toggle

Enjoy2Live:
- installed @radix-ui/react-switch
- added helpful css color utilities
- added data-[disabled] state
- allow react-docgen-typescript to pull @radix-ui declarations
- feat(components): Added Switch

### DIFF
--- a/packages/UI/src/components/Switch/Switch.stories.tsx
+++ b/packages/UI/src/components/Switch/Switch.stories.tsx
@@ -1,0 +1,45 @@
+import type { Meta, StoryObj } from '@storybook/react';
+
+import Switch from './Switch';
+
+const meta: Meta<typeof Switch> = {
+  component: Switch,
+};
+
+export default meta;
+type Story = StoryObj<typeof Switch>;
+
+/*
+ *👇 Render functions are a framework specific feature to allow you control on how the component renders.
+ * See https://storybook.js.org/docs/react/api/csf
+ * to learn how to use render functions.
+ */
+export const Untoggled: Story = {
+  render: () => <Switch />,
+  parameters: {
+    design: {
+      type: 'figma',
+      url: 'https://www.figma.com/file/rvdnSOeHdjODE0U7xdNTR9/DL-Universal-Design-System-(NEW)?type=design&node-id=12-3844&mode=design&t=8VGSbFcJwjPcf02d-4',
+    },
+  },
+};
+
+export const Toggled: Story = {
+  render: () => <Switch defaultChecked />,
+  parameters: {
+    design: {
+      type: 'figma',
+      url: 'https://www.figma.com/file/rvdnSOeHdjODE0U7xdNTR9/DL-Universal-Design-System-(NEW)?type=design&node-id=12-3838&mode=design&t=8VGSbFcJwjPcf02d-4',
+    },
+  },
+};
+
+export const Disabled: Story = {
+  render: () => <Switch disabled />,
+  parameters: {
+    design: {
+      type: 'figma',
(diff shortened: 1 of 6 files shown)


packages/UI/src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,62 @@
...
+ * @see http://localhost:6006/?path=/docs/components-switch--docs
+ * @see https://www.radix-ui.com/primitives/docs/components/switch#root
+ */
+const Switch = React.forwardRef<
+  React.ElementRef<typeof SwitchPrimitives.Root>,
+  SwitchProps
+>(({ className, ...props }, ref) => (
+  <SwitchPrimitives.Root
+    className={cn(
+      'peer inline-flex h-6 w-12 shrink-0 cursor-pointer items-center rounded-2xl outline outline-1 -outline-offset-1 outline-grayscale-800 transition-colors disabled:cursor-not-allowed disabled:!bg-grayscale-100 disabled:!outline-GreyScale-charcoal data-checked:bg-grayscale-800 data-unchecked:bg-grayscale-50',
```
l1keafox, line 51: Because of the outline-1 don't you need a m-1?

Also what does the ! do in "disabled:!bg-grayscale-100" ?

packages/UI/src/components/Switch/Switch.stories.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const meta: Meta<typeof Switch> = {
+  component: Switch,
+};
+
+export default meta;
+type Story = StoryObj<typeof Switch>;
+
+/*
+ *👇 Render functions are a framework specific feature to allow you control on how the component renders.
```
l1keafox, line 13: Add Documentation here about Using an Label and htmlFor w/ id to link the two. 
One thing I noticed about both radix/and Shadcn/ui is that when you click the label, it also triggers a switch.

packages/UI/src/components/Switch/Switch.tsx
Enjoy2Live: why do you think I need m-1?

! in "disabled:!bg-grayscale-100" adds important to that rule.

https://tailwindcss.com/docs/configuration#important-modifier

packages/UI/src/components/Switch/Switch.stories.tsx
Enjoy2Live: good point, I'm having a [discussion with Anouar](https://www.figma.com/file/rvdnSOeHdjODE0U7xdNTR9?node-id=12:3800&mode=dev#697460378) on whether we should add a label component or not in Figma, great minds think alike!

htmlFor w/ id is something that should be known by developers generally since it's not custom made by us but it's still a good idea to add it once we have a `WithLabel` story.

packages/UI/src/components/Switch/Switch.tsx
l1keafox: I was mistaken, it was with "outline-2" that you need "m-1."
Outline is outside the element, it will look misaligned at a certain point and needs margins in order to align correctly.

l1keafox: One other question is why are we using outline here instead of border?

Enjoy2Live: > One other question is why are we using outline here instead of border?

because I need to use it as "inner border" that's why I have accompanied it with  `-outline-offset-1` which basically puts the outline 1px inside the content box

Enjoy2Live: outline offset is something you can't do with border but you have to know that outline is also not calculated as part the box-model size where border is calculated.

PR #1676, opened 2024-02-11
fix(components): typescript errors in PositionCard

Enjoy2Live:


Enjoy2Live:
@judeamos this is the only change I'd make to your `development/components`. lemme know what you think

PR #1678, opened 2024-02-11
Development/user profile

Enjoy2Live:


Enjoy2Live:
this needs to be `process.env.NAME` rather than `process.env.NODE_ENV` as the latter only returns Production or Development and never returns "staging"
https://github.com/dev-launchers/dev-launchers-platform/blob/3a25af0f289e75250aa25a0849bdf7b84da63085/apps/user-profile/src/utils/featureFlags.js#L1

Enjoy2Live:
@judeamos why do we have apps/user-profile/src/context/UserDataContext.js? it makes more sense if we use one context at a time, all teams use packages/UI/src/context/UserDataContext.js

Enjoy2Live:
> @judeamos why do we have apps/user-profile/src/context/UserDataContext.js? it makes more sense if we use one context at a time, all teams use packages/UI/src/context/UserDataContext.js

using these two different contexts can cause issues because we're making two requests at the same time

PR #1685, opened 2024-02-20
Components/lucide react iconography

lamodots:


packages/UI/src/docs/System/Iconography.mdx
```diff
@@ -13,6 +12,17 @@ import * as Icons from '../../assets/icons';
...
+
+## Lucide Icons
+
+<IconGallery>
+  {Object.entries(icons).map(([IconName, IconJSX]) => (
+    <IconItem name={IconName} key={IconName}>
+      <IconJSX />
+    </IconItem>
+  ))}
 </IconGallery>
```
Enjoy2Live, line 27: This is a good start but to make it easier to navigate maybe we should put lucide react icons in a collapsible section with guide on how to import them.

same thing to be said about custom-made icons.

something like this

https://components-iconography-suggestion--632ca21aeea1618b7fa0bf48.chromatic.com/?path=/docs/iconography--docs

code that made it happen https://github.com/dev-launchers/dev-launchers-platform/pull/1688

Enjoy2Live (changes requested):
This is a good start but to make it easier to navigate maybe we should put lucide react icons in a collapsible section with guide on how to import them.

Enjoy2Live (approved):
LGTM, this will be included in our next release. Thanks alot for providing the docs @lamodots .

it's time to see if members will use icons properly now!

PR #1686, opened 2024-02-20
Components/1655dropdownmenu

l1keafox:


Enjoy2Live (changes requested):
Looks good, maybe we need to add a form example in the future because it's getting a bit hard to know how to use those with a form especially for those who aren't familiar with html forms and always use react libraries to submit forms.

Please go over the requested changes and let me know if you have questions.
Thanks as always @l1keafox

packages/UI/src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -0,0 +1,240 @@
+'use client';
+
+import * as DropdownMenuPrimitive from '@radix-ui/react-dropdown-menu';
+import { Checkbox } from 'components/Checkbox';
```
Enjoy2Live, line 4: this should be `../Checkbox`

Enjoy2Live (approved):
LGTM, thanks for adding the form example.
at some point we need to write a story dedicated to forms and with something (React Hook Form + zod probably) that makes the form validated and less boring to write

PR #1690, opened 2024-02-24
community/try vike with styled components

Enjoy2Live:
- Wrap the whole app with ThemeProvider
- unify styled-components versions in order to hoist to one shared dependency
- remove useless wrapping style tag


Enjoy2Live:
> Oh, so removing the style tag in the head for the onRenderHtml.tsx and instead just calling the sheet.getStyleTags() worked?
> 
> Is it because the sheet.getStyleTags() gets the style tags as part of the function? I'd assume so since it says it in the function name.

Exactly but having wrapping style wasn't the ultimate issue but the versions inconsistency was.

If you bring back the wrapping style you'll notice that it works but upon rendering the component the styles take a second to appear

PR #1691, opened 2024-02-25
Development/user profile

judeamos:
Changes made: 
- I've removed user data context from the user-profile folder, now we're using the global context in the packages folder. 
- Reflected changes where needed for using global `UserDataContext.js`
- cleaned up code from merging. 
- Changes from master branch have been merged into this one as well.


apps/app/next.config.js
```diff
@@ -7,7 +7,6 @@ const withTM = require('next-transpile-modules')([
   '@devlaunchers/site-projects',
   '@devlaunchers/dev-recruiters',
   '@devlaunchers/website',
-  '@devlaunchers/user-profile',
```
Enjoy2Live, line 10: why remove this? I think it's needed

Enjoy2Live: maybe it was part of a conflict but definitely needed to be back

Enjoy2Live (changes requested):

apps/app/next.config.js
judeamos: ` '@devlaunchers/user-profile'` is duplicated, so i deleted one of them. If you expand it you'll see it on top

Enjoy2Live (approved):
LGTM

PR #1695, opened 2024-02-29
feat(workspaces): lint staged files before comitting

### LINKED ISSUE #1661: Run eslint to lint staged files before they get committed
So far the only way to fix lint issues that are auto fixable is by running `yarn workspace @devlaunchers/nameOfApp fix-all-files` or save the files using vscode `Ctrl + S`. we can't expect all members to remember to lint their apps before committing so it's a good idea to use something like [lint-staged](https://www.npmjs.com/package/lint-staged) to do this automatically for new commits.

## Acceptance Criteria

- [ ] **Scenario 1: eslint on staged files before committing**
  - Given the project has @devlaunchers/eslint-config-bases installed as dev dependency with its configurations
  - And there are staged files ready to be committed
  - When a commit is made
  - Then lint-staged should run eslint on the staged files
  - And any lint issues should be auto-fixed if possible
  - And the commit should proceed only if there are no lint issues remaining in the staged files

- [ ] **Scenario 2: Lint issues are not auto-fixed before committing**
  - Given the project is set up with linting rules and @devlaunchers/eslint-config-bases configuration
  - And there are staged files with non-auto-fixable lint issues ready to be committed
  - When a commit is made
  - Then lint-staged should run eslint on the staged files
  - And the commit should fail if there are any remaining lint issues in the staged files

Enjoy2Live:
This PR adds a git hook that ensures whenever commit is made, the commit follows the code style devlaunchers uses (so far prettier only) by formatting the committed files then actually commit them.

This should help reduce the massive merge conflicts like ones that include an empty space only
closes #1661


Enjoy2Live:
just realized this breaks the build system, fixing now!

PR #1714, opened 2024-03-07
Icons documentation

lamodots:
Updated the documentation for both Custom icons and Lucide React icons. 
Showed.
- How to use and modify Lucide Icons
- How to use, modify and create custom icons. 
For custom icons, I will like  have more explanation on how to use the props mentioned in the UD documentation on icons CHARACTERESTICS so it could be covered in the documentation. Currently all the custom icons we have available do not follow the UD doc on Icons.


Enjoy2Live:
closing this as we already have https://github.com/dev-launchers/dev-launchers-platform/pull/1685

PR #1733, opened 2024-03-24
Staging/ideaspace

dbradham:
This pull request is to deploy the latest changes from IdeaSpace project


Enjoy2Live (changes requested):
@dbradham you need to run `yarn install` otherwise the build won't succeed as the deps have changed.

When running it it should make changes to yarn.lock, add those changes to this PR and it should make the build actions pass.

dbradham:
> @dbradham you need to run `yarn install` otherwise the build won't succeed as the deps have changed.
> 
> When running it, it should make changes to yarn.lock, add those changes to this PR and it should make the build actions pass.

Thanks. This should be addressed now, but would be good to have you review again.

Enjoy2Live (approved):
Looks good to me, congrats on getting closer to a new release!

RETprojects (approved):
This PR looks good to merge into production. Great job, everyone!

PR #1740, opened 2024-03-28
1701 Added Sidebar base structure in user profile

tiwarineelu59:


Enjoy2Live:
Good to see teams using tailwind now, I was wondering why you created 
`apps/user-profile/src/components/modules/UserProfile/SideBar/StyledSideBar.js`
File? It seems redundant as you're styling with tailwind, let's remove this file

PR #1754, opened 2024-04-01
Resolve TypeScript and Styling Conflicts with Radix and Tailwind

NAsriGhada:
I'm facing an error when using Radix UI components in our Tailwind-configured project. After migrating from JSX I got a different error, attached in the screenshot, which suggests a type incompatibility with the Progress components as far as I understood.

Inline styles were a successful interim solution for the initial Tailwind conflict, but this TypeScript issue persists.

Screenshot 2024-04-01 204516


Enjoy2Live:
@NAsriGhada this PR is huge and is so different from `development/components`, can you describe to me how you ended making the PR step by step?

PR #1770, opened 2024-04-14
Staging/ideaspace

dbradham:


apps/ideaspace/src/components/modules/DashboardPage/DashboardPage.js
```diff
@@ -25,21 +24,28 @@ function DashboardPage() {
...
+        if (item.author.data !== null) {
+          item.author = cleanData(item.author.data);
+        }
 
-        setLoading(false);
-        setSourceCards(allCards);
+        return {
+          ...item,
+          mostRecentCommentTime: new Date(
+            item.comments[0]?.updated_at
```
chungthuang, line 42: Do we need to search through all comments?

apps/ideaspace/src/components/modules/EditIdea/EditIdea.js
```diff
@@ -82,15 +77,17 @@ function EditIdea() {
 
   useEffect(async () => {
     if (ideaId) {
-      const idea = cleanData(await agent.Ideas.getIdea(ideaId, new URLSearchParams("populate=*")));
+      const idea = cleanData(
+        await agent.Ideas.getIdea(ideaId, new URLSearchParams('populate=*'))
```
chungthuang, line 81: @becevka we should revisit how many collections are being queried with `populate=*`.

chungthuang (approved):

apps/ideaspace/src/components/modules/EditIdea/EditIdea.js
dbradham: > @becevka we should revisit how many collections are being queried with `populate=*`.

@chungthuang @becevka how would we access info on the author of the idea without populating at least one "level"?

apps/ideaspace/src/components/modules/DashboardPage/DashboardPage.js
dbradham: > Do we need to search through all comments?

@chungthuang it's a requirement to count the total number of comments for a given user.

chungthuang: Sorry my question was if we have a list of comments, don't we need to iterate through the list to find the most recent comment time? The current logic just check the first comment. Or is the API response already sorted by updated  time?

apps/ideaspace/src/components/modules/EditIdea/EditIdea.js
chungthuang: I think you can get it from `populate=author`. But I didn't mean to block this PR, just something we were discussing in the backend team about the complexity of the relations between our collections.

dbradham: > I think you can get it from `populate=author`. But I didn't mean to block this PR, just something we were discussing in the backend team about the complexity of the relations between our collections.

That's good to know. Thanks!

apps/ideaspace/src/components/modules/DashboardPage/DashboardPage.js
dbradham: Oh, I see. We can test to make sure the API response is sorted by updated time. Otherwise I think you are correct and we need to make an update.

PR #1814, opened 2024-06-12
1813 tailwind styling

tiwarineelu59:


packages/tailwind-constructor/tailwind.config.js
```diff
@@ -24,6 +24,10 @@ module.exports = withTV({
       transparent: 'transparent',
       ...filterTokensByType('color', tokens),
     },
+    extend: {
+      colors: {
+        'sky-blue': '#7dd3fc',
+      }},
```
Enjoy2Live, line 30: @Nouri-Anouar can we approve this?

Nouri-Anouar: @Enjoy2Live This color does not exist in our design system. We can make an exception, but I need to review the design first to see where it is being used.

PR #1815, opened 2024-06-12
1813 tailwind styling

tiwarineelu59:


judeamos (approved):
Approving changes, but can you delete the styledIntroductionmodal.js and have the styling applied inline.

PR #1833, opened 2024-06-19
Development/user profile

tiwarineelu59:


packages/utility/agent.ts
```diff
@@ -1,75 +1,50 @@
-import { NewApplicant, Opportunity, Project, User as UserType, Idea, Like, Save } from "@devlaunchers/models";
```
Enjoy2Live, line 1: can we revert the changes that removed comments from this file `packages/utility/agent.ts`? why have you removed it in the first place if I may ask?

Enjoy2Live (changes requested):
Let's abstain from making gigantic PRs to master so we can easily review and give feedback.

I reviewed the parts that are relevant to me only as I can't go over all the changes, please read the requested change and let me know what you think here.

PR #1866, opened 2024-07-15
roles filter bug fix

ajaywayase:


lamodots (changes requested):
role.attribute is repetitive . You can do object  destructuring

ajaywayase:
> role.attribute is repetitive . You can do object destructuring

Can you check now

lamodots (approved):

PR #1875, opened 2024-07-20
Staging/user profile — July 28th Release

judeamos:
### User Profile code changes for July 28th Release.

Changes Made: 
- UI revamp for our use profile redesign.
- Major code cleanup within our user profile project. (includes renaming, deleting, and restructuring of files)
- Usage of Tailwind in new components, and the refactoring of old components to use Tailwind.
- The onboarding process has been added with backend integration, this includes:
1. A new "onboarding" page access through `<base-url>/onboarding` 
2. A new Onboarding Modal that shows up if the user has not completed the onboarding process
- Loading functionality, a spinner is shown whenever the user data is being requested. For good user feedback. 
- We've added our own personal context files for our project and management of our state management system.
— You can find this change in the context folders
— I've built a very simple redux like system for our project (without needing to install the library). You can find these changes in the `user-profile/src/state/...` folder where i'm using reducers and actions. 
— If you'd like more understanding of this implementation, you can watch my [loom video tutorial](https://www.loom.com/looms/videos/State-Management-873e9a9dfb9a42d4a56a75f0b65c7dbb) i created for my team 

**Where Review is mostly needed**
- In setting up tailwind — made changes in `packages/tailwind-contructor/tailwind.config.js`
- In implementing the profile upload functionality (being able to change your profile) the `agent.ts` file was modified in `packages/utility/agent.ts` 
— Changes were also made to fix conflicts from the master branch.

- Added loading & error state in the `packages/UI/src/context/UserDataContext.js` (used for showing spinner in the user profile component)
- We've added multiple images to the image folder, not sure if this is gonna be an issue in terms of performance. We've  mostly stuck to svg files, and some png/jpeg files.  @Enjoy2Live 
- I've moved the sign up page redirection from the `src/components/modules/UserProfile/UserProfile.js` into the `src/pages/users/me.js` & added it to the `src/pages/onboarding.js`  (onboarding page). 
— Not sure who added the logic but wanted to let you know @Enjoy2Live @chungthuang. 

- I've used the `setUserData()` method in the `user-profile/src/components/modules/UserProfile/EditProfileModal/SettingPage/SettingPage.js` (**Line 54**), can you check if i'm using it correctly @Enjoy2Live @Nouri-Anouar.  
— This is for updating the users bio details in the backend. 
— The frontend user data state would not update unless you refreshed it to pull changes from the backend. 
— So i've just update the state: `setUserData({ ...userData, bio: editProfileState.bio })`
—  "...userData" gets existing state, then just updating the bio property.

[...]

PR #1878, opened 2024-07-20
quick change back (mixed package.json)

judeamos:
I made changes in the wrong package.json file in this PR (https://github.com/dev-launchers/dev-launchers-platform/pull/1877). I've now undone the changes i made in the wrong file. Again, it is better to review in this PR (https://github.com/dev-launchers/dev-launchers-platform/pull/1875)

### DIFF
--- a/apps/user-profile/package.json
+++ b/apps/user-profile/package.json
@@ -78,6 +78,10 @@
     "not ie <= 11",
     "not op_mini all"
   ],
+  "lint-staged": {
+    "*.js": "eslint '*/**/*.{js,jsx}' --fix",
+    "*.{js,css,md}": "prettier --write"
+  },
   "resolutions": {
     "styled-components": "^5"
   },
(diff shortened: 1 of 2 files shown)


apps/user-profile/package.json
```diff
@@ -78,6 +78,10 @@
     "not ie <= 11",
     "not op_mini all"
   ],
+  "lint-staged": {
+    "*.js": "eslint '*/**/*.{js,jsx}' --fix",
+    "*.{js,css,md}": "prettier --write"
+  },
```
Enjoy2Live, line 84: @judeamos 
this should be deleted, also please don't ignore code-owners alert and always seek their review before merging

judeamos: I was reverting the change back to the original, as I accidentally changed the one in the universal design folder. So the package.json in the user profile folder already had this. Thought there was no need for review as this was there by default. Where changes are being made outside of our project folder, or any app-level changes I'll make sure to include it in the description, just like I've done here: (https://github.com/dev-launchers/dev-launchers-platform/pull/1875/files)

judeamos: Do you need me to remove this, even though this was their originally?

judeamos: Oh i've just checked the commits log, it was, my bad. Reverting it back.

PR #1883, opened 2024-07-21
Lucas community

LucasWyski001:
added projects feature within the navigation which then allows users to click there and be redirected to the projects page, as well as implemented code for the projects page to load and render images without error


Enjoy2Live (approved):
LGTM but let's make sure other teams aren't bothered with these changes

dbradham:
@LucasWyski001 can you pull the latest changes from `master` into `development/community` ? I think that would remove a lot of the proposed changes here within the ideaspace app.

PR #1890, opened 2024-07-27
added the notification popover

yacineKahlerras:


Enjoy2Live (changes requested):
Very good starting point, I would suggest moving `packages/UI/src/components/molecules/NotificationPopover/*`
to `packages/UI/src/components/organisims/Navigation/NotificationPopover/*`

Also please rebase from `development/notification` to get access to notification page and minor fixes I made to global styles.
```
git rebase origin/development/notification
git push -f
```

PR #1894, opened 2024-08-04
fix(Need to trigger click function for upvote button)

RETprojects:


dbradham (approved):

dbradham:
@Enjoy2Live we made some changes here to the agent.ts and would appreciate your review-

The solution for delete was something I found in this thread-https://stackoverflow.com/questions/51069552/axios-delete-request-with-request-body-and-headers

I was not able to pass in a body even with this change, and I ended up just adding it to the end of the request string to get this to work for us. Not sure if you've used the delete endpoints much, or see any issues with /likes/delete implementation in agent.ts, but something seems strange there to me.

Enjoy2Live:
> @Enjoy2Live we made some changes here to the agent.ts and would appreciate your review-
> 
> The solution for delete was something I found in this thread-https://stackoverflow.com/questions/51069552/axios-delete-request-with-request-body-and-headers
> 
> I was not able to pass in a body even with this change, and I ended up just adding it to the end of the request string to get this to work for us. Not sure if you've used the delete endpoints much, or see any issues with /likes/delete implementation in agent.ts, but something seems strange there to me.

I won't be able to look at the code on my computer until tomorrow but http specification explicitly states that the DELETE method does not expect a request body so I don't see it as a good idea to go against you specs

Enjoy2Live:
@dbradham I think upgrading axios would solve this issue, newer versions allow payload with http DELETE method

dbradham:
> @dbradham I think upgrading axios would solve this issue, newer versions allow payload with http DELETE method

Thanks, that makes sense. I'm going to avoid upgrading axios because we have a working solution here, and that represents a lot of extra risk with only 1 day before code freeze.

PR #1904, opened 2024-08-18
feat(footer): add a contact email

Enjoy2Live:


chungthuang:
Can you share a screenshot of how this will look like?

chungthuang (approved):

zar92:
I would also like to see how it actually looks like. The contact email should be before Terms of Service.
Screenshot 2024-08-15 184902

Enjoy2Live:
@chungthuang @zar92 
image

PR #1906, opened 2024-08-18
Staging/ideaspace

dbradham:


chadcrotchett (approved):
👍

dbradham:
@Enjoy2Live @dev-launchers/dev-recruit @judeamos @dev-launchers/user-profile @dev-launchers/platform-devs hi all, we are hoping to get this merged to staging for testing ahead of tomorrow's code freeze.

Please let us know if you have any feedback. Thanks!

apps/user-profile/src/components/common/Icons/SVG/FilledCross.js
chungthuang, line 1: This is changed by the linter?

packages/utility/agent.ts
```diff
@@ -82,7 +82,7 @@ const requests = {
   patch: <T>(url: string, body: {}) =>
     axios.patch<T>(url, body).then(responseBody),
   delete: <T>(url: string, body?: {}) =>
-    axios.delete<T>(url, body).then(responseBody),
+    axios.delete<T>(url, { data: body }).then(responseBody),
```
chungthuang, line 85: I think this is because strapi added the data field when we migrated from v3 to v4. Should we use an empty object if body is empty though? I think this code will send `{"data":{}}` instead of `{}`. 
Let's do it in another PR: does that mean put and patch should also wrap the body under `data`?

apps/user-profile/src/components/common/Card/Card.js
```diff
@@ -2,7 +2,8 @@ import Header from './Header';
 function Card({ title, children, onEdit }) {
   return (
     <div className="flex flex-col shadow-lg w-full h-full bg-white">
-      <Header title={title} onEdit={onEdit} /> <div className="px-8 pb-10 pt-6">{children}</div>
+      <Header title={title} onEdit={onEdit} />{' '}
```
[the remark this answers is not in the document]
dbradham: @judeamos please review

apps/user-profile/src/components/common/Icons/SVG/FilledCross.js
dbradham: @chungthuang yes I believe so. I did not touch this file and do not see why Remi would have touched it either.

packages/utility/agent.ts
dbradham: @chungthuang I discussed this with Mohammed on a previous pull request (https://github.com/dev-launchers/dev-launchers-platform/pull/1894)

To summarize, the body could not reach the backend before or after this change, but I left it like this because I found a thread that said the way we have implemented previously is wrong-https://stackoverflow.com/questions/51069552/axios-delete-request-with-request-body-and-headers

Mohammed proposed upgrading Axios to solve this issue. I think that is outside the scope of this ticket as we are able to send the DELETE requests with expected results now.

package.json
```diff
@@ -63,6 +63,7 @@
     "@commitlint/config-conventional": "17.0.3",
     "@devlaunchers/eslint-config-bases": "workspace:^",
     "@playwright/test": "1.24.2",
+    "@types/node": "22.2.0",
```
Enjoy2Live, line 66: Why adding this dep in the global package.json? It should be added to its respective workspace otherwise errors might appear globally

Enjoy2Live (changes requested):

apps/user-profile/src/components/modules/Onboarding/UserOnboardingModal/PlatformOnboarding/PageFive/StyledPageFive.js
```diff
@@ -35,8 +35,6 @@ export const UserInterestsContainer = styled.div`
   flex-wrap: wrap;
   row-gap: 20px;
   column-gap: 24px;
-  max-height: 308px;
-  overflow: auto;
```
judeamos, line 39: This should not be deleted. Please revert.

PR #1944, opened 2024-09-01
Staging/ideaspace

dbradham:


RETprojects (approved):

dbradham:
@dev-launchers/platform-enablement when you get a chance, we would appreciate your review here!

packages/UI/src/components/atoms/Radio/StyledRadio.ts
```diff
@@ -16,13 +16,15 @@ export const Option = styled.label`
...
     pointer-events: none;
   }
+
+  margin-bottom: 0.25rem;
 `;
 
 export const Input = styled.input`
   height: 24px;
-  width: 24px;
+  /* width: 24px; */
```
chungthuang, line 25: Why is this commented out?

apps/ideaspace/src/components/modules/SubmissionForm/SubmissionForm.js
```diff
@@ -32,13 +32,13 @@ function SubmissionForm() {
   );
 
   const getDBInvolveLevel = (involveString) => {
-    if (involveString.includes('and also believe')) {
+    if (involveString.includes('to end')) {
```
chungthuang, line 35: This looks hacky. Can we change the backend to store this as an enum?

apps/ideaspace/src/components/common/IdeaForm/IdeaForm.js
```diff
@@ -7,227 +7,236 @@ import SubmissionButton from './SubmissionButton';
...
-				enableReinitialize
-			>
-				{({ errors, setFieldValue, touched }) => (
-					<Form>
-						<AutoSubmitToken />
-						<atoms.Typography type='h4'>
-							Idea Info
-							<hr noshade="false" style={{margin: "1rem auto 2rem"}}/>
-						</atoms.Typography>
+  const compareValuesToInitial = (values) => {
```
chungthuang, line 28: Can you give an example of what are we comparing?

apps/ideaspace/src/components/common/IdeaForm/IdeaForm.js
```diff
@@ -7,227 +7,236 @@ import SubmissionButton from './SubmissionButton';
...
+                    />
+                    <atoms.Radio
+                      onChange={(e, label) =>
+                        setFieldValue('involveLevel', label)
+                      }
+                      label='I want to "own" this idea only after it is approved as a project'
+                    />
+                    <atoms.Radio
+                      onChange={(e, label) =>
+                        setFieldValue('involveLevel', label)
```
chungthuang, line 183: Can we set field to "highly", so we don't need this conversion https://github.com/dev-launchers/dev-launchers-platform/pull/1944/files#r1744700300?

packages/UI/src/components/atoms/Radio/StyledRadio.ts
chadcrotchett: hard-coded pixel width was causing the radio 'circles' to be 'ovals' that were stretched out

apps/ideaspace/src/components/modules/SubmissionForm/SubmissionForm.js
dbradham: @chungthuang  We can, but seems out of scope for this frontend change. We would still need to map the values to an enum.

Do you have a solution in mind that avoids making a comparison like this to get a value to save to the database?

dbradham: @chungthuang just saw your other comment and updated this

apps/ideaspace/src/components/common/IdeaForm/IdeaForm.js
dbradham: @chungthuang yes thanks for the recommendation here, this is complete.

apps/ideaspace/src/components/common/IdeaForm/IdeaForm.js
dbradham: @chungthuang this is to check if any form values have been updated.

apps/ideaspace/src/components/modules/WorkshoppingPage/CommentsForm/SingleComment.js
```diff
@@ -139,12 +139,12 @@ function SingleCommentComponent(props) {
             </div>
           </SingleCommentContent>
           <SingleCommentButtons style={{ display: 'flex' }}>
-            <LikeButton
+            {/* <LikeButton
```
chungthuang, line 142: Why are we commenting this out?

chungthuang (approved):

PR #1962, opened 2024-09-11
Addition of projects tab on the landing page & loading of projects page correctly.

Aishwaryak01:
Addition of projects tab on the landing page & loading of projects page correctly.


packages/UI/src/components/molecules/Alert/Alert.tsx
dbradham, line 1: this component is currently marked as deprecated, so may behave erratically in some contexts due to legacy styling. it should be fine to use, but I just wanted to call out the fact that it is deprecated in the storybook.

packages/UI/src/components/organisms/NavDropdown/NavDropdown.tsx
```diff
@@ -83,12 +83,12 @@ const NavDropdown = ({
               );
             }
             return (
-              <>
+              <React.Fragment key={i}>
```
dbradham, line 86: what is the purpose of this change?

yarn.lock
```diff
@@ -2970,7 +2970,7 @@ __metadata:
   resolution: "@devlaunchers/site-projects@workspace:apps/site-projects"
   dependencies:
     "@babel/core": ^7.18.9
-    "@devlaunchers/eslint-config-bases": "workspace:^"
+    "@devlaunchers/eslint-config-bases": "workspace:*"
```
dbradham, line 2973: what is the purpose of this change?

PR #1966, opened 2024-09-12
fix(Disabled vote button when not logged in)

RETprojects:


dbradham (approved):
looks good. We should get to the bottom of the authenticated user issue

PR #2051, opened 2024-10-02
Sorting dropdown

vivienzhou2014:
https://github.com/user-attachments/assets/03722809-85d2-40e2-8ad3-1fd016af5a38

1. Fixed the width not consistent issue for hover box in dropdown options
2. Added white background in toggle when toggle is clicked to open the options, changed options border radius setting to make toggle and options more unite. Added shadow to the whole component when toggle is clicked to open to fit the requirement in [figma](https://www.figma.com/design/hEDIHsBZYa6hmK8As5mcjy/%5BDL---IdeaSpace%5D-UX-designs?node-id=11680-17769&node-type=instance&m=dev).
3. Changed font size to 1rem in toggle and 0.875rem in options. @dbradham Please let me know if it is ok to touch the Styled.Typography.ts file. If not, I will try to find another way to adjust the font size for options.

### DIFF
--- a/apps/ideaspace/src/components/modules/BrowseIdeas/StyledBrowseIdeas.js
+++ b/apps/ideaspace/src/components/modules/BrowseIdeas/StyledBrowseIdeas.js
@@ -38,4 +38,5 @@ export const FilterDiv = styled.div`
   display: flex;
   flex-direction: column;
   align-items: flex-end;
+  margin-bottom: 4rem;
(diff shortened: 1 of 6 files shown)


AlinaU1 (approved):

dbradham (approved):
Only comment is that ideally we could standardize how we apply styles. But that is out of scope here.

Issue #1269, opened 2023-07-10
[US - DEV] As MILES HARDIN, I want to be able to navigate back from the 'terms and conditions' page to the previous page I visited.


JulieMass:
### Considerations:
Miles Hardin wants to be able to navigate back from the 'terms and conditions' page to the previous page he visited. 

Please refer to:
- UX designs in [Figma ](https://www.figma.com/file/hEDIHsBZYa6hmK8As5mcjy/%5BDL---IdeaSpace%5D-UX-designs?node-id=2645%3A5947&t=YzCPmKYI4hbrBbcZ-1 ) 

### Acceptance criteria:
- When Miles Hardin clicks on the 'back' button on the 'terms and conditions' page, he should be taken back to the previous page he visited.
- The URL of the previous page should be preserved.
- The previous page should be displayed in the same state as when Miles Hardin left it.
- Layout is web responsive and the functionality were tested thoroughly and works seamlessly on all devices and [target browsers ](https://docs.google.com/spreadsheets/d/1upYP1mvHkU3sRBg4gCeGG4P63tO5-whn6VKLMlr6qYI/edit?usp=sharing ): Chrome, Microsoft Edge, Safari, Firefox and Opera.

----------------------------------------------------
### Sub-tasks:
- [X] UXR
- [X] UX design
- [x] Development
- [x] QA
- [x] Bug: If use link from Summary table, back button navigates to previously clicked Section ("Age Eligibility", "Changes to Terms", etc)
----------------------------------------------------

![image.png](https://images.zenhubusercontent.com/63400f3d1e79614e45d9ebd9/d8e7d151-3a12-4bd6-a1bd-749dff2d6803)

JulieMass:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/IdeaSpace-Team-6350185d9843157427cc900b/issues/dev-launchers/dev-launchers-platform/1269?planning-poker) with Zenhub @alicefate @dbradham @jordanmcousins @JulieMass @katehirschman @MariliaDP @RETprojects

JulieMass:
Hi @alicefate ,

As we discussed yesterday in our sprint planning, I’ve created this ticket for the back button for the terms and conditions. Please let me know if the description and the acceptance criteria of the ticket are OK and/or if I missed anything.

Thank you,

dbradham:
@JulieMass This is blocked until #750 is completed per the acceptance criteria.

[...]

PR #2057, opened 2024-10-06
Ahsaask/terms condition

Ahsaask:
- Fixed UI for Terms and condition based on figma 
- Auto save in ideaform: using local-storage  
- fixed some minor bugs in idea submission form
- added back-button to terms and condition
- added logic to back button to exit tab when no back url available
- added responsiveness to terms and condition

Issues: #1269 #1952 

@dbradham @chadcrotchett @AlinaU1


staging/kustomization.yaml
Ahsaask, line 1: I did not change this file, not sure why its here. (possibly came when i merged it with development/ideaspace)

dbradham: @Ahsaask looks like this is from the linter tool, it just added some whitespace and changed double quotes to single quotes. Should be fine to merge.

apps/ideaspace/src/components/common/BackButton/BackButton.js
```diff
@@ -8,8 +8,10 @@ const BackButton = ({ buttonType, clickHandler, backRoute }) => {
   const goBack = () => {
     if (backRoute) {
       router.push(backRoute);
+    } else if (window.history.length > 1) {
+      window.history.back();
     } else {
-      window.history.back(-1);
+      window.close();
```
dbradham, line 14: is closing the current window when users navigate part of the requirements for this change? @Ahsaask @chadcrotchett

Ahsaask: Requirement: I want to be able to navigate back from the 'terms and conditions' page to the previous page I visited.

We decided that when the user navigates to a new tab for idea-space terms condition, we will close the tab to prompt user to get back to the submission form. Ideally this should not replace any other functionalities. Let me know if there are some concerns.

staging/kustomization.yaml
chadcrotchett: @dbradham Is it an option to 'linter' the whole code base in one PR so these whitespace formatting items don't distract future PRs?

PR #2069, opened 2024-10-13
feat(design): updates to size, space and font token sets

annaisaacks:
Defines primitive text values:

- font-family
- font-weight
- font-style


dbradham:
@annaisaacks is this change related to any specific ticket on the universal design board? I am asking so I can confirm these changes have the expected result before we merge to staging

annaisaacks:
@Nouri-Anouar 

Do we have a ticket for adding variables?

Nouri-Anouar:
Hi @dbradham 
Here is a ticket for this PR #2087 

@annaisaacks Please review and lmk if I missed something.

dbradham (approved):
Approving with the understanding that the tokens file will be transferred to universal design for code ownership moving forward.

PR #2072, opened 2024-10-13
/ideaspace/workshop/{id} uses only usernames for both the owner and the commentor

yacineKahlerras:
-it shows the idea owner by their username not their email
-shows commentor by their username not their full name


dbradham (approved):
Looks good!

PR #2073, opened 2024-10-15
Ahsaask/terms condition fixes

Ahsaask:
- fixed back button bug
- fixed back button placement 
- fixed margin, gap & spacing issues in Terms & Condition
- Hide back button for mobile devices
- managed breakpoints for different screen sizes

#1269 #1952 

@dev-launchers/ideaspace-team


chadcrotchett (approved):
👍

PR #2083, opened 2024-10-20
Onboarding Modal Refinement

judeamos:
Huge refinement of the onboarding modal.

**Changed Made**
- Transitioned all components under `src/components/modules/UserOnboardingModal`  to use tailwind, except `PageSix` component which will take some thinking as it uses css animations.
- Added new component based icons `common/Icons/SVG/FilledCheckmark.js` 
- Modified 2 existing icon components have `<button>` parents to allow `onClick`  and other button attributes
- Code clean up - deleted or moved images that were not needed anymore. 
- Code refactoring on all components within the `src/components/modules/UserOnboardingModal`  as well.


anne-creator (approved):
Wow, it looks like the tons of change. I checked the user profile overview page and it looks good, except the Avatar in the header bar and the person's profile photo don't have a default photo setting if I didn't upload my photo.

If it is the case before switching to the tailwind css then we are good.

Screenshot 2024-10-20 at 10 33 02 PM

PR #2089, opened 2024-10-24
1993 remove generate idea button

Ahsaask:
- Removed generate idea button
- New style change and responsiveness


chadcrotchett (approved):
yep

PR #2126, opened 2024-11-10
Landing page/div container

nandbolt:
- created a div container component that supports vertical, horizontal and wrap that has an optional reverse setting
- implemented stories for the component


dbradham:
Closing this PR as @Nouri-Anouar confirmed it is no longer needed.

PR #2147, opened 2024-12-06
Migrate Button from styled components to tailwind

yacineKahlerras:


dbradham (approved):
looks good, thanks for removing the "Deprecated" tag as well!

PR #2150, opened 2024-12-07
migrating nav link to new design

yacineKahlerras:
- converted nav-links component to tailwind
- fixed type errors that are related to the Button component
- created Notification model
- fixed type errors related to Notification Popover

### DIFF
--- a/packages/UI/src/components/atoms/NavLink/NavLink.stories.tsx
+++ b/packages/UI/src/components/atoms/NavLink/NavLink.stories.tsx
@@ -1,7 +1,22 @@
-import type { Meta } from '@storybook/react';
+import type { Meta, StoryObj } from '@storybook/react';
 import NavLink from './NavLink';
 
 export default {
-  title: 'Deprecated/Atoms/NavLink',
+  title: 'components/atoms/NavLink',
   component: NavLink,
 } as Meta<typeof NavLink>;
+
+type Story = StoryObj<typeof NavLink>;
+
+export const Default: Story = {
+  args: {
+    children: 'Contact Us',
+  },
+};
+
+export const Disabled: Story = {
+  args: {
+    disabled: true,
+    children: 'Contact Us',
+  },
+};
--- a/packages/UI/src/components/organisms/Navigation/NotificationPopover/NotificationPopover.tsx
+++ b/packages/UI/src/components/organisms/Navigation/NotificationPopover/NotificationPopover.tsx
@@ -3,9 +3,13 @@ import { Popover, PopoverContent, PopoverTrigger } from '../../../Popover';
 import { Bell, Settings } from 'lucide-react';
 import NotificationItem from '../../../NotificationItem/NotificationItem';
 import { agent } from '@devlaunchers/utility';
+import { Notification } from '@devlaunchers/models';
+
+const SettingsIcon = Settings as React.FC<React.SVGProps<SVGSVGElement>>;
+const BellIcon = Bell as React.FC<React.SVGProps<SVGSVGElement>>;
 
 export default function NotificationPopover() {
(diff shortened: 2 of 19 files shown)


packages/models/notification.tsx
```diff
@@ -0,0 +1,43 @@
+import * as React from 'react';
+import Avatar from '../UI/src/components/atoms/Avatar/Avatar';
+
+export type NotificationUser = {
+  data: {
+    id: number;
+    attributes: {
+      username?: string;
```
dbradham, line 8: @yacineKahlerras is there any specific reason we are marking this as optional?

packages/UI/src/components/organisms/Navigation/NotificationPopover/NotificationPopover.tsx
```diff
@@ -112,3 +119,6 @@ export default function NotificationPopover() {
     </Popover>
   );
 }
+
+const SettingsIcon = Settings as React.FC<React.SVGProps<SVGSVGElement>>;
```
dbradham, line 123: @yacineKahlerras what do you think of putting these lines at the top of the file, for readability?

packages/UI/src/components/atoms/NavLink/NavLink.stories.tsx
```diff
@@ -1,7 +1,22 @@
-import type { Meta } from '@storybook/react';
+import type { Meta, StoryObj } from '@storybook/react';
 import NavLink from './NavLink';
 
 export default {
-  title: 'Deprecated/Atoms/NavLink',
+  title: 'components/atoms/NavLink',
```
dbradham, line 5: thanks for this!

packages/models/notification.tsx
yacineKahlerras: @dbradham yeah for some of the accounts they ca be null, so i left it as an option to maybe put an email address if the field doesn't exist  do you want me to set the type as `username: string | null` instead ?

packages/UI/src/components/organisms/Navigation/NotificationPopover/NotificationPopover.tsx
yacineKahlerras: @dbradham yeah that works, its just type casting because typescript was complaining so i figured they didnt matter as much

packages/UI/src/components/organisms/Navigation/NotificationPopover/NotificationPopover.tsx
```diff
@@ -3,9 +3,15 @@ import { Popover, PopoverContent, PopoverTrigger } from '../../../Popover';
 import { Bell, Settings } from 'lucide-react';
```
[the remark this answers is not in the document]
yacineKahlerras: @dbradham no they're being used as icons, this `Settings as React.FC<React.SVGProps<SVGSVGElement>>;` is just type casting solution for this typescript error : 

`'Settings' cannot be used as a JSX component.
  Its type 'LucideIcon' is not a valid JSX element type.
    Type 'ForwardRefExoticComponent<LucideProps>' is not assignable to type '(props: any, deprecatedLegacyContext?: any) => ReactNode'.
      Type 'ReactElement<any, string | JSXElementConstructor<any>> | null' is not assignable to type 'ReactNode'.
        Type 'ReactElement<any, string | JSXElementConstructor<any>>' is not assignable to type 'ReactNode'.
          Property 'children' is missing in type 'ReactElement<any, string | JSXElementConstructor<any>>' but required in type 'ReactPortal'.ts(2786)
index.d.ts(205, 9): 'children' is declared here.`

dbradham (approved):

PR #2153, opened 2024-12-12
fix(button): fixed prop errors for button component

yacineKahlerras:
added some props validation to avoid any future errors


dbradham (approved):
looks good

PR #2165, opened 2025-01-03
feat (pages): Redesign Landing Page

annaisaacks:


packages/UI/src/components/organisms/Footer/Footer.jsx
```diff
@@ -1,81 +1,107 @@
...
+      {/* Bottom Section */}
+      <div className="border-t border-gray-700 bg-black">
+        <div className="max-w-7xl mx-auto px-6 py-4 flex flex-col md:flex-row items-center justify-between">
+          {/* Links */}
+          <div className="flex space-x-4 text-sm">
+            <a
+              href="#"
+              className="hover:underline text-purple-400 hover:text-purple-500"
+            >
+              © 2024 Dev Launchers, Inc.
```
reginaldojunior, line 64: Hello! 

I think here can use current year already (2025) 😄  

What you think?

annaisaacks: Updated, thank you!

Issue #1254, opened 2023-07-07
As a potential member, I want to be able to upload a file for resume and other attachments from local device


Srividya1394:
- In the volunteer application form, add Upload file button under the "Please include your portfolio/resume" field
- On clicking the upload file button, it should display dialog box with "CHOOSE FILES" where user can choose file from device
- Once file is chosen, the dialog box will display the file attached along with options "REMOVE FILE" and "ADD MORE FILES" as buttons
- See the images below for reference
![image.png](https://images.zenhubusercontent.com/6460df08ff04bb3b6e33b654/98ddc516-7e9c-41e9-bc01-ebfc60cef82a)
![image.png](https://images.zenhubusercontent.com/6460df08ff04bb3b6e33b654/1e753d80-5aeb-4b40-8404-177a014073bc)

PR #2166, opened 2025-01-06
To upload a file for resume and other attachments from local device

jeyanthi-sm:
In the volunteer application form, there is an 'Upload File' button. When the button is clicked, a modal window opens with the text 'CHOOSE FILES/Drop file,' allowing the user to select a file from their device. After a file is selected, there should be an option to delete it and upload a new file if the wrong file was chosen. 
https://github.com/dev-launchers/dev-launchers-platform/issues/1254

### DIFF
--- a/apps/dev-recruiters/src/components/modules/FormPage/signUpForm.tsx
+++ b/apps/dev-recruiters/src/components/modules/FormPage/signUpForm.tsx
@@ -1,10 +1,17 @@
+'use client';
 import { atoms, organisms } from '@devlaunchers/components/src/components';
 import FormErrorScroller from '@devlaunchers/components/src/utils/formErrorScroller';
 import { Opportunity } from '@devlaunchers/models';
 import { NewApplicant } from '@devlaunchers/models/newApplicant';
 import { agent } from '@devlaunchers/utility';
-import { Field, Formik, FormikHelpers } from 'formik';
-import { MouseEventHandler, useState } from 'react';
+import {
+  Field,
+  Formik,
+  FormikHelpers,
+  useFormikContext,
+  useField,
+} from 'formik';
+import { MouseEventHandler, useEffect, useState } from 'react';
 import { ThemeProvider } from 'styled-components';
 import theme from '@devlaunchers/dev-recruiters/src/styles/theme';
 import * as Yup from 'yup';
@@ -13,7 +20,23 @@ import {
   CloseButton,
   CloseIcon,
 } from '../DetailedPage/PositionCard/StyledPositionCard';
-import { GradientLine, SubmitButton } from './styledSignupForm';
+import {
+  CancelUploadButton,
+  GradientLine,
+  ModalUploadSection,
+  SubmitButton,
+  UploadButton,
+} from './styledSignupForm';
+import UploadModal from './uploadModal';
+import DragAndDrop from '../NewJoinPageComponent/Drag and Drop';
+import axios from 'axios';
+import Modal from '../DetailedPage/PositionPopupModal';
+
+interface UploadProps {
+  handleUploadCloseModal?: () => void;
+  handleOkCloseModal?: () => void;
+  handleCancelCloseModal?: () => void;
+}
 
 interface FormFields extends Omit<NewApplicant, 'level'> {
   level: NewApplicant['level'] | '';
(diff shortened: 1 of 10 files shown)


chungthuang:
There's a merge conflict in `apps/dev-recruiters/src/components/modules/FormPage/signUpForm.tsx`.

apps/dev-recruiters/src/components/modules/NewJoinPageComponent/Drag and Drop/draganddrop.tsx
```diff
@@ -0,0 +1,225 @@
...
+        ...prevState,
+        e.dataTransfer.files[0],
+      ]);
+    }
+    fileUpload(e.dataTransfer.files);
+  }
+
+  function dragLeaver(e: any): void {
+    e.preventDefault();
+    //e.stoppropagation();
```
dbradham, line 115: is there a specific reason to leave this commented code in place? could it be deleted?

apps/dev-recruiters/src/components/modules/NewJoinPageComponent/Drag and Drop/draganddrop.tsx
```diff
@@ -0,0 +1,225 @@
...
+        );
+        setIsUploading(false);
+        uploadStatus = false;
+        return false;
+      }
+      // Check file size
+
+      if (inFiles[0].size > maxSizeInBytes) {
+        setUploadError(
+          'The file you have chosen is not a valid file type. Please upload only .doc, .pdf, .jpg, .jpeg, and .png (Max 25MB)'
```
dbradham, line 54: could we abstract this to a constant and reference the constant on line 44 also? That would make it easier to maintain, since it seems this is the same message as displayed on line 44.

apps/dev-recruiters/src/components/modules/NewJoinPageComponent/Drag and Drop/draganddrop.tsx
```diff
@@ -0,0 +1,225 @@
...
+    setDragActive(false);
+  }
+
+  function dragEnter(e: any): void {
+    e.preventDefault();
+    //e.stoppropagation();
+    setDragActive(true);
+  }
+
+  function dragOverHandler(e: any): void {
```
dbradham, line 125: could this be combined with the function on line 119? it looks like they are doing the same thing.

dbradham:
@jeyanthi-sm this looks good, I just had a few minor comments. Also, it looks like there are some conflicts on the signUpForm if you could please resolve those.

One question here is if you are familiar with Tailwind? That may make some of the styles here easier to maintain going forward. Not a requirement to use Tailwind, but it is something other teams are implementing.

apps/dev-recruiters/src/components/modules/FormPage/signUpForm.tsx
```diff
@@ -83,6 +157,53 @@ export default function SignUpForm({
...
+    setFilesUploaded([]);
+    setShowUploadModal(false);
+  };
+
+  const handleRemoveFile = () => {
+    setIsDeleting(true); // Deleting state
+    const newArr = [...selectedFiles];
+    axios
+      .delete(
+        `${process.env.NEXT_PUBLIC_STRAPI_URL}/googledrive/${filesUploaded['id']}`
```
chungthuang, line 191: We try to keep all the logic to query backend in https://github.com/dev-launchers/dev-launchers-platform/blob/master/packages/utility/agent.ts

apps/dev-recruiters/src/components/modules/FormPage/signUpForm.tsx
```diff
@@ -83,6 +157,53 @@ export default function SignUpForm({
...
+      )
+
+      .then((responseBody) => {
+        if (responseBody.status === 200) {
```
chungthuang, line 195: Should we show an error if the file couldn't be deleted?

apps/dev-recruiters/src/components/modules/NewJoinPageComponent/Drag and Drop/draganddrop.tsx
```diff
@@ -0,0 +1,225 @@
...
+      }
+      return true;
+    };
+    const responseResult = await response();
+
+    if (responseResult === true) {
+      portfolioUploadformData.append('files', inFiles[0]);
+
+      const postResult = await axios
+        .post(
```
chungthuang, line 69: Same suggestion https://github.com/dev-launchers/dev-launchers-platform/pull/2166/files#r1921576475.

packages/utility/agent.ts
```diff
@@ -71,6 +71,9 @@ function createFormData(item: any) {
 const responseBody = (response: AxiosResponse) =>
   response.data.data ? response.data.data : response.data;
 
+const deleteStatus = (response: AxiosResponse) =>
+  response.status === 200 ? response.status : response.status;
```
chungthuang, line 75: This seems like a no-op? I think it always returns `response.status`?

packages/utility/agent.ts
```diff
@@ -82,10 +85,11 @@ const requests = {
   patch: <T>(url: string, body: {}) =>
     axios.patch<T>(url, body).then(responseBody),
   delete: <T>(url: string, body?: {}) =>
-    axios.delete<T>(url, body).then(responseBody),
-  postForm: (url: string, data: FormData) =>
+    //  axios.delete<T>(url, body).then(responseBody),
+    axios.delete<T>(url, body).then(deleteStatus),
```
chungthuang, line 89: So we are only returning status, not the response body?

packages/models/newApplicant.ts
```diff
@@ -9,8 +9,9 @@ export interface NewApplicant {
     skill: string;
   }[];
   commitment: number;
-  extraInfo?: string;
-  portfolioLink?: string;
+  extraInfo?: string; //column missing
```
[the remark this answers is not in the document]
chungthuang: Is the column still missing?

chungthuang (approved):

packages/utility/agent.ts
jeyanthi-sm: The response body contains the record details if the ID is present. However, from these details, it is not possible to determine whether the delete operation succeeded or not. So, I am returning the status.

chungthuang:
I squashed the commits into 1.

PR #2224, opened 2025-03-06
Fixed the dialog box and improved Checkbox UI

Abd-pathan:
Fixed the dialog box and improved Checkbox UI, and refined form's validation sequence.
Basically we first check for the required fields in the form and then we check for the checkbox.


AlinaU1 (approved):
👍

Issue #2237, opened 2025-03-19
I want to see who created an idea in the idea card


AlinaU1:
As an IdeaSpace user, when browsing ideas, I want to see the idea owner's name on the idea card before clicking on it.

**Acceptance Criteria:**
- [x] The idea owner's name and profile image are displayed on each idea card.
- [x] The name and profile image are correctly linked to the corresponding user’s data.
- [x] The displayed information matches the data in the user's profile and Workshopping page.

**Now:** 
<img width="449" alt="Image" src="https://github.com/user-attachments/assets/5ac407b9-848d-4d09-abbe-fc13df589101" />

**Updated**:

<img width="629" alt="Image" src="https://github.com/user-attachments/assets/a296f692-bd91-4228-88b1-3d4173af6610" /> <img width="525" alt="Image" src="https://github.com/user-attachments/assets/8810d853-1bad-4add-8ee9-a8f1c9a7e635" />

Figma: https://www.figma.com/design/hEDIHsBZYa6hmK8As5mcjy/%5BDL---IdeaSpace%5D-UX-designs?node-id=11044-17402&t=1I9xafEPZkMQslgv-4

PR #2241, opened 2025-03-23
Display author info in the idea card

ravindrasinghshah:
[We want to see who created an idea in the idea card](https://github.com/dev-launchers/dev-launchers-platform/issues/2237)

**Description:**
* Updated BrowseIdeas.js to populate query with deep data of author for profile picture and display name.
* Updated IdeaCard.js to show author profile picture and author display name.
    * If author display name is empty then fallback to username
* Used Avatar atom to display author profile picture.
    *  If profile picture is not available then avatar fallbacks to user-profile svg.

**Screenshots:**
Before:
image

After:
image


dbradham (approved):
build failure is unrelated to these changes which look good. Thanks for using the atoms!

PR #2275, opened 2025-04-06
Dbradham/type script updates

dbradham:


packages/UI/src/types/MobileNavigationDropdownItem.ts
```diff
@@ -0,0 +1,4 @@
+export default interface ProfileDropdownProps {
```
srushti-ghatage, line 1: Can you rename this object name to be same as the file name?

dbradham: Updated!

dbradham:
@srushti-ghatage I made the requested changes here, when you get a chance to review again.

Nouri-Anouar (approved):

chungthuang (approved):

PR #2316, opened 2025-04-20
fixed typing errors

srushti-ghatage:


yarn.lock
```diff
@@ -11362,6 +11363,15 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@types/node@npm:22.14.1":
```
dbradham, line 11366: Is this package necessary to fix the errors?

packages/utility/package.json
```diff
@@ -13,5 +13,8 @@
     "@devlaunchers/models": "workspace:^",
     "axios": "^0.27.2",
     "eslint": "8.21.0"
+  },
+  "devDependencies": {
+    "@types/node": "22.14.1"
```
dbradham, line 18: Is this needed to solve the type errors?

dbradham:
Looks like we are not getting a successful build before running the type checks here. This may be related to adding additional packages to the yarn config.

Screenshot 2025-04-20 at 11 44 14 AM

packages/utility/package.json
srushti-ghatage: @dbradham 
Yes, the type errors were being thrown for "process.env" which is a node object. One of the solutions I found for this was to add the above line in the package.json. Also, it's added as a dev dependency hence should not be a problem I believe. Yet I can check for why the build is failing.

yarn.lock
srushti-ghatage: Replied here: https://github.com/dev-launchers/dev-launchers-platform/pull/2316/files#r2051778321

packages/utility/package.json
srushti-ghatage: I have reverted back the code yet I am seeing those errors.
I have reached out to @chungthuang already for the same.

chungthuang:
@srushti-ghatage can you rebase to resolve the conflicts?

PR #2379, opened 2025-05-11
chore(codeowner): change codeowner to anyone on the platform-devs team

chungthuang:
As we transition into a wider development department, developers will often work on apps in different projects


dbradham (approved):
Thanks for getting this updated!

PR #2384, opened 2025-05-15
Victor l idea card changes 

liangz25:
tile, end with "..."


.github/CODEOWNERS
chungthuang, line 1: Looks like `development/ideaspace` need to be rebased on the master branch. This file was recently updated on the master branch to https://github.com/dev-launchers/dev-launchers-platform/blob/master/.github/CODEOWNERS.

Dockerfile
```diff
@@ -140,7 +140,7 @@ USER nextjs
 
 EXPOSE ${NEXTJS_APP_PORT:-3000}
 
-ENV NEXT_TELEMETRY_DISABLED 1
```
chungthuang, line 143: Where you testing with Docker?

production/kustomization.yaml
```diff
@@ -2,9 +2,9 @@ apiVersion: kustomize.config.k8s.io/v1beta1
...
-- wordpress.yaml
+  - ../workload
+  - wordpress.yaml
 images:
-- name: devlaunchers/platform-website
-  newName: devlaunchers/platform-website
-  newTag: 2.24.0 # {"$imagepolicy": "platform-website:platform-website:tag"}
+  - name: devlaunchers/platform-website
+    newName: devlaunchers/platform-website
+    newTag: 2.24.0 # {"$imagepolicy": "platform-website:platform-website:tag"}
```
ravindrasinghshah, line 10: Looks like file formatted with space is committed in this PR, maybe we should revert it back.

.github/CODEOWNERS
ravindrasinghshah: done

Dockerfile
ravindrasinghshah: Looks like this came because of wrong base branch, post rebase with master issue is resolved.

production/kustomization.yaml
dbradham: Not a big deal... but would speed things up to revert this to skirt around code owners blocking PR from being merged.

PR #2393, opened 2025-05-29
240/user not published

dbradham:
This allows users to access the onboarding page after they create an account.

This fixes the issues with users not having profiles.

This also fixes some form validation issues with the save button.

### DIFF
--- a/apps/user-profile/src/components/modules/Onboarding/UserLandingPage/UserLandingPage.js
+++ b/apps/user-profile/src/components/modules/Onboarding/UserLandingPage/UserLandingPage.js
@@ -58,7 +58,7 @@ export default function UserLandingPage() {
       errors.isFirstNameValid = false;
       errors.firstNameError = 'First name is not valid';
     } else {
-      errors.isFirstNameValid = formValue.firstName.length > 0;
+      errors.isFirstNameValid = formValue.firstName.trim().length > 0;
     }
 
     if (!formValue.lastNameTouched) {
@@ -68,14 +68,14 @@ export default function UserLandingPage() {
       errors.isLastNameValid = false;
       errors.lastNameError = 'Last name is not valid';
     } else {
-      errors.isLastNameValid = formValue.lastName.length > 0;
+      errors.isLastNameValid = formValue.lastName.trim().length > 0;
     }
 
     if (!formValue.displayNameTouched) {
       errors.displayNameError = 'Display name is required';
       errors.isDisplayNameValid = false;
     } else {
-      errors.isDisplayNameValid = formValue.displayName.length > 0;
+      errors.isDisplayNameValid = formValue.displayName.trim().length > 0;
     }
 
     return errors;
@@ -162,21 +162,18 @@ export default function UserLandingPage() {
               { withCredentials: true }
             )
             .then(() => {
-              router
-                .push({
-                  pathname: '/users/me',
-                })
-                .then(() => {
-                  setSaveInProgress(false);
-                });
+              router.push({
+                pathname: '/users/me',
+              });
             })
             .catch((error) => {
-              setSaveInProgress(false);
               console.log('Error Updating User Data: ', error);
             });
         })
         .catch((error) => {
           console.log('Error Updating Profile Data: ', error);
+        })
+        .finally(() => {
           setSaveInProgress(false);
         });
     } else {
--- a/apps/user-profile/src/components/modules/Onboarding/UserLandingPage/UserLandingPage.js
+++ b/apps/user-profile/src/components/modules/Onboarding/UserLandingPage/UserLandingPage.js
@@ -1,6 +1,7 @@
 import React, { useState, useEffect } from 'react';
 import { useRouter } from 'next/router';
 import agent from '@devlaunchers/utility/agent';
+import axios from 'axios';
 import CameraIcon from './../../../../../src/images/camera-icon.svg';
 import Button from '@devlaunchers/components/components/atoms/Button';
 import { useUserDataContext } from '@devlaunchers/components/context/UserDataContext';
@@ -33,9 +34,8 @@ export default function UserLandingPage() {
   const { userData } = useUserDataContext();
   const [person, setPerson] = useState({ ...initialFormValue });
   const [saveInProgress, setSaveInProgress] = useState(false);
-  const [formValidation, setFormValidation] = useState(
-    validate(initialFormValue)
-  );
+  const [formValidation, setFormValidation] = useState({});
+  const [termsAccepted, setTermsAccepted] = useState(false);
 
   const [profilePicture, setProfilePicture] = useState(
     userData?.profile?.profilePictureUrl
@@ -58,7 +58,7 @@ export default function UserLandingPage() {
       errors.isFirstNameValid = false;
       errors.firstNameError = 'First name is not valid';
     } else {
-      errors.isFirstNameValid = true;
+      errors.isFirstNameValid = formValue.firstName.length > 0;
     }
 
     if (!formValue.lastNameTouched) {
@@ -68,21 +68,15 @@ export default function UserLandingPage() {
       errors.isLastNameValid = false;
       errors.lastNameError = 'Last name is not valid';
     } else {
-      errors.isLastNameValid = true;
+      errors.isLastNameValid = formValue.lastName.length > 0;
     }
 
     if (!formValue.displayNameTouched) {
       errors.displayNameError = 'Display name is required';
       errors.isDisplayNameValid = false;
     } else {
-      errors.isDisplayNameValid = true;
+      errors.isDisplayNameValid = formValue.displayName.length > 0;
     }
-
-    errors.isFormValid = () =>
-      formValue.termsAndConditions &&
-      errors.isFirstNameValid &&
-      errors.isLastNameValid &&
-      errors.isDisplayNameValid;
 
     return errors;
   }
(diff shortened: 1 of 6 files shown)


RETprojects (approved):

apps/user-profile/src/pages/onboarding.js
```diff
@@ -33,16 +33,12 @@ export default function UserOnboardingPage() {
...
-  //   else if (!isLoading && isAuthenticated && hasAcceptedTermsOfService) {
-  //     router.push('/users/me');
-  //   }
-  // }, [isAuthenticated]); // Run the effect when `isAuthenticated` changes
+  useEffect(() => {
+    // Redirect to the user profile page if terms of service are accepted
+    if (!isLoading && isAuthenticated && hasAcceptedTermsOfService) {
+      router.push('/users/me');
+    }
+  }, [isAuthenticated]); // Run the effect when `isAuthenticated` changes
```
ravindrasinghshah, line 41: Small nit - useEffect might want other variables in dependency as well.

apps/user-profile/src/components/modules/UserProfile/Overview/Overview.js
```diff
@@ -16,6 +17,15 @@ import Interests from './Interests';
...
   const { userData } = useUserDataContext();
+  const router = useRouter();
+
+  useEffect(() => {
+    if (!userData.hasAcceptedTermsOfService) {
+      router.push({
+        pathname: '/onboarding',
+      });
+    }
+  }, []);
```
ravindrasinghshah, line 28: Small nit - useEffect might want `userData.hasAcceptedTermsOfService` in dependency.

apps/user-profile/src/components/modules/Onboarding/UserLandingPage/UserLandingPage.js
```diff
@@ -68,22 +68,16 @@ export default function UserLandingPage() {
...
-      errors.isLastNameValid = true;
+      errors.isLastNameValid = formValue.lastName.length > 0;
     }
 
     if (!formValue.displayNameTouched) {
       errors.displayNameError = 'Display name is required';
       errors.isDisplayNameValid = false;
     } else {
-      errors.isDisplayNameValid = true;
+      errors.isDisplayNameValid = formValue.displayName.length > 0;
```
ravindrasinghshah, line 78: I am not sure if we want to cover whitespaces validation in length - if we do then maybe trim().length can be used.

apps/user-profile/src/components/modules/UserProfile/Overview/Overview.js
dbradham: @ravindrasinghshah what would this help with? The goal here is to re-rout the user if they access this page and have no accepted the ToC. I only want this check to execute once when the page loads, so I am not adding any in dependency.

apps/user-profile/src/components/modules/Onboarding/UserLandingPage/UserLandingPage.js
dbradham: @ravindrasinghshah great point! I am adding this to all of the checks here.

apps/user-profile/src/components/modules/Onboarding/UserLandingPage/UserLandingPage.js
```diff
@@ -143,33 +137,43 @@ export default function UserLandingPage() {
...
+              router
+                .push({
+                  pathname: '/users/me',
+                })
+                .then(() => {
+                  setSaveInProgress(false);
+                });
+            })
+            .catch((error) => {
+              setSaveInProgress(false);
```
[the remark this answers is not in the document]
dbradham: @ravindrasinghshah great catch! I am updating this now.

apps/user-profile/src/components/modules/Onboarding/UserLandingPage/UserLandingPage.js
```diff
@@ -33,9 +34,8 @@ export default function UserLandingPage() {
   const { userData } = useUserDataContext();
   const [person, setPerson] = useState({ ...initialFormValue });
   const [saveInProgress, setSaveInProgress] = useState(false);
-  const [formValidation, setFormValidation] = useState(
-    validate(initialFormValue)
-  );
+  const [formValidation, setFormValidation] = useState({});
```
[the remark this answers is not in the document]
dbradham: This was the original implementation, but we had a lot of unstable behavior with validate being called more than expected, so we simplified things.

apps/user-profile/src/components/modules/Onboarding/UserLandingPage/UserLandingPage.js
```diff
@@ -143,33 +137,43 @@ export default function UserLandingPage() {
...
 
     if (userId) {
       const profileRequestBody = {
-        data: {
-          user: userData?.id,
-          displayName: `${person.displayName}`,
-          profilePictureUrl: uploadedProfilePicture || profilePicture,
-          hasAcceptedTermsOfService: true,
-        },
+        user: userData?.id,
```
chungthuang, line 145: We already have `userId` defined in line 141.

apps/user-profile/src/components/modules/Onboarding/UserLandingPage/UserLandingPage.js
```diff
@@ -224,7 +228,7 @@ export default function UserLandingPage() {
             </div>
           </div>
         </div>
-        <form className="flex flex-col">
+        <div className="flex flex-col">
```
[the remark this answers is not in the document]
dbradham: We do not want the page to reload, so this was removed.

apps/user-profile/src/components/modules/Onboarding/UserLandingPage/UserLandingPage.js
```diff
@@ -143,33 +137,43 @@ export default function UserLandingPage() {
...
         .then(() => {
-          setSaveInProgress(false);
+          const userRequestBody = {
+            hasAcceptedTermsOfService: true,
+          };
 
-          router.push({
-            pathname: '/users/me',
-          });
+          axios
```
chungthuang, line 158: Should we move this method into [agent](https://github.com/dev-launchers/dev-launchers-platform/blob/master/packages/utility/agent.ts)?

dbradham: Ideally yes, but I think that is out of scope here. This was the original code we brought back after realizing it is necessary. (this code still running in production)

Right now, the agent doesn't have a put method for `User`, and will require more testing to implement.

apps/user-profile/src/components/modules/UserProfile/Overview/Overview.js
```diff
@@ -16,6 +17,15 @@ import Interests from './Interests';
 function Overview() {
   // Access user data from the context
   const { userData } = useUserDataContext();
+  const router = useRouter();
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
dbradham: See comment below.

apps/user-profile/src/pages/onboarding.js
```diff
@@ -33,16 +33,12 @@ export default function UserOnboardingPage() {
...
-  //   // Redirect to signup if authenticated but no username is set
-  //   if (!isLoading && isAuthenticated && userData.name === '') {
-  //     router.push('/signup');
-  //   }
-  //   // Redirect to the user profile page if terms of service are accepted
-  //   else if (!isLoading && isAuthenticated && hasAcceptedTermsOfService) {
-  //     router.push('/users/me');
-  //   }
-  // }, [isAuthenticated]); // Run the effect when `isAuthenticated` changes
+  useEffect(() => {
```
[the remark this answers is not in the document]
dbradham: `isLoading` will update at the same time as `isAuthenticated`, so not sure there is any value to adding that.

`hasAcceptedTermsOfService` will only ever change when this form is submitted, so also not sure it makes any sense to include that here.

chungthuang (approved):

apps/user-profile/src/components/modules/UserProfile/Overview/Overview.js
ravindrasinghshah: I agree - we often come across such use cases where we think adding all deps is not necessary, which usually is true...as in this case. But more or less it's to abide with the react's design philosophy and practical considerations. Some of the points I keep in my thoughts are:

1. React's rules of hooks rely on declarative and predictable behavior, when we break those. there are build warnings.
2. If you reference a variable inside useEffect, and don’t list it as a dependency, the value inside the effect may become stale - which often result in bugs or harder to debug.
3. Skipping dependencies can cause effects to run with stale or unexpected values - even if we know today that some variable "won’t change," someone else might not. Or change logic in the future, and our assumption breaks.
4. Standardize coding standards - by having consistent implementation behavior, reviews, debugging gets easier.

apps/user-profile/src/components/modules/Onboarding/UserLandingPage/UserLandingPage.js
j-peace: Just for context.
It was necessary to change because before this change, all of users were mocked as `hasAcceptedTermsOfService` = `true`.
Is that rigth?
Is there another reason for this refactor?

apps/user-profile/src/components/modules/Onboarding/UserLandingPage/UserLandingPage.js
```diff
@@ -354,15 +356,14 @@ export default function UserLandingPage() {
                 Cancel
               </Button>
               <Button
-                type="submit"
                 buttonType="secondary"
                 buttonSize="xl"
                 onClick={onContinueClick}
-                disabled={!formValidation.isFormValid() || saveInProgress}
+                disabled={!formValidation.isDisplayNameValid || !termsAccepted}
```
j-peace, line 362: Is it necessary to keep the Button disable while `saveInProgress`?

apps/user-profile/src/components/modules/Onboarding/UserLandingPage/UserLandingPage.js
dbradham: This change is to ensure we update the user object in the database to indicate they accepted the ToC.

We also need to redirect them to /users/me after the data is sent to the backend.

PR #2413, opened 2025-06-05
recover lost changes

dbradham:


apps/user-profile/src/components/modules/Onboarding/UserLandingPage/UserLandingPage.js
```diff
@@ -133,38 +121,35 @@ export default function UserLandingPage() {
...
-        },
+        user: userId,
+        displayName: person.displayName,
+        profilePictureUrl: uploadedProfilePicture || profilePicture,
       };
+
       setSaveInProgress(true);
 
-      agent.Profiles.put(profileId, profileRequestBody)
+      agent.Profiles.post(profileRequestBody)
```
chungthuang, line 152: Is this going to create a new profile?

chungthuang:
Can you change the commit message to `fix(user profile): recover lost changes`

apps/user-profile/src/components/modules/Onboarding/UserLandingPage/UserLandingPage.js
dbradham: Yes, we should use `post` to create a new object.

image

dbradham:
> Can you change the commit message to `fix(user profile): recover lost changes`

Just made a new commit with this message @chungthuang

chungthuang (approved):

apps/user-profile/src/components/modules/Onboarding/UserLandingPage/UserLandingPage.js
chungthuang: Sorry I mean to say, can this lead to us creating multiple profiles for the same user? Does the code handle create profile if it doesn't exist, but update if there's already one.

PR #2415, opened 2025-06-08
feat: deprecated StyledCheckbox, added missing hover ring

srushti-ghatage:
# Screenshots

## Before
Screenshot 2025-06-08 015539

## After
Screenshot 2025-06-08 014159


apps/dev-recruiters/src/components/common/Checkbox/CheckBox.tsx
```diff
@@ -1,13 +1,5 @@
 import React from 'react';
-import styled from 'styled-components';
 import { CheckboxContainer, CheckboxInput, CheckMark } from './StyledCheckbox';
-// import {
-//   CheckboxContainer,
-//   HiddenCheckbox,
-//   StyledCheckbox,
-//   Icon,
-//   Label,
-// } from './StyledCheckbox';
```
srushti-ghatage, line 3: This component is still being used inside of CheckboxList component which is being used at multiple places.
Screenshot 2025-06-08 020042

In order to completely get rid of the StyledCheckbox (using styled components), this would be the last effort.

srushti-ghatage: @dbradham Let me know if I should get that removed in this same PR.

srushti-ghatage: cc: @chungthuang

dbradham: Thanks for calling this out @srushti-ghatage !

I would suggest to make a new pull request for these changes.

Ideally we could delete this file entirely so no devs use it in the future. (increasing the chances they use the shared component checkbox you updated)

srushti-ghatage: Created another PR for this effort -> https://github.com/dev-launchers/dev-launchers-platform/pull/2416

There could be conflict once the current PR is merged, but I can take a look at it later once merged.

dbradham (approved):

PR #2423, opened 2025-06-12
fix: Incorrect Redirect on Idea Card 

Abd-pathan:
Updated the IdeaCard component to redirect to the workshopping page when clicking anywhere on the card.


AlinaU1 (approved):
👍

PR #2580, opened 2025-06-22
Upstream/development/recruitment

dbradham:


packages/models/dltalcommuser.ts
```diff
@@ -0,0 +1,8 @@
+// DLTALCOMMUNITY USER MODEL
+
+export interface DlTalCommUser {
+  name: string;
+  email: string;
+  skills: string;
+  roles: string;
```
j-peace, line 7: The user can have more than 1 role? It should be a string?

apps/dev-recruiters/next.config.js
```diff
@@ -45,12 +45,17 @@ const nextConfig = {
...
+      'devlaunchersstaging.blob.core.windows.net',
+      'lh3.googleusercontent.com',
+      'localhost',
+      'apiv4-staging.devlaunchers.org',
+      'apiv4.devlaunchers.org',
     ],
     disableStaticImages: true,
   },
-  webpack5: true,
+  // webpack5: true, next 12 later versions use webpack 5 by default.
```
j-peace, line 58: What is menas

j-peace (changes requested):

packages/models/dltalcommuser.ts
dbradham: @j-peace this is correct per the current backend schema:

image

apps/dev-recruiters/src/components/modules/TalcommunityPage/TalFormPage/signUpForm.tsx
```diff
@@ -1,13 +1,14 @@
 import { atoms, organisms } from '@devlaunchers/components/src/components';
 import FormErrorScroller from '@devlaunchers/components/src/utils/formErrorScroller';
 import { agent } from '@devlaunchers/utility';
-import { Field, Form, Formik, FormikHelpers } from 'formik';
+import { Field, Form, Formik, FormikHelpers, useFormik } from 'formik';
```
j-peace, line 4: This `useFormik` has been added. But it is not being used. Right?

apps/site-projects/next.config.js
```diff
@@ -59,3 +63,9 @@ const nextConfig = {
   },
 };
 module.exports = withPlugins([[imagesPlugin], nextConfig]);
+
+// REMOVE Unsupported Properties:
+// webpack5: true, // No longer needed
+// target: "server", // Deprecated
+// configOrigin: "default", // Not valid
+// webpackDevMiddleware: (config) => config, // Remove this unless you need it
```
j-peace, line 71: This comment could be deleted?

j-peace (approved):

apps/site-projects/next.config.js
dbradham: I didn't author any of this, but think it may possibly be useful in the future, so want to leave it for the time being, although it's definitely a little ugly.

PR #2599, opened 2025-07-06
Typography updates

dbradham:
- Recent updates to use the shared Typography component in the application's main pages


dbradham:
> @dbradham 
> 
> The code looks good!! And after push this to staging, what are the steps to reproduce?
> 
> I am wondering in things like Modal and Alert that are more sensive of changes. And if it will still working as expected.
> 
> 

Good question. There are many pages and components impacted here, so it would be best for us to have a couple of people scan each page of the website (only 10-15 total pages) and ensure things look as expected. Most of the code here was authored by Harsh and I just helped solve some conflicts and build failures. So he should have (hopefully) done some preliminary tests prior to making the PR. We will need extra eyes to validate this!

PR #2624, opened 2025-07-23
Ideaspace alerts

Ahsaask:
- add reusable alert component
- fix redirect issue in submission form
- success/error pop ups for both edit and submission forms


apps/ideaspace/src/components/common/SubmissionAlert/Alert.js
```diff
@@ -0,0 +1,183 @@
...
+  type = 'success',
+  variant = 'submit',
+  onClose,
+  countdown = null,
+}) => {
+  const [currentCountdown, setCurrentCountdown] = useState(countdown);
+
+  // Auto-dismiss logic for edit alerts
+  useEffect(() => {
+    if (variant === 'edit') {
```
ravindrasinghshah, line 13: To improve the readability and maintenance of code - variant values should be defined and read from enum instead of hardcoding.

apps/ideaspace/src/components/common/SubmissionAlert/Alert.js
```diff
@@ -0,0 +1,183 @@
+import React, { useState, useEffect } from 'react';
+
+const Alert = ({
+  type = 'success',
```
ravindrasinghshah, line 4: In code logic below type is used as a Boolean - Maybe it makes sense to use bool itself instead of type. I think success flow should be higher than error, we can have `isSuccess = true`

apps/ideaspace/src/components/common/SubmissionAlert/Alert.js
```diff
@@ -0,0 +1,183 @@
...
+          xmlns="http://www.w3.org/2000/svg"
+        >
+          <path
+            d="M22 11.0799V11.9999C21.9988 14.1563 21.3005 16.2545 20.0093 17.9817C18.7182 19.7088 16.9033 20.9723 14.8354 21.5838C12.7674 22.1952 10.5573 22.1218 8.53447 21.3744C6.51168 20.6271 4.78465 19.246 3.61096 17.4369C2.43727 15.6279 1.87979 13.4879 2.02168 11.3362C2.16356 9.18443 2.99721 7.13619 4.39828 5.49694C5.79935 3.85768 7.69279 2.71525 9.79619 2.24001C11.8996 1.76477 14.1003 1.9822 16.07 2.85986M9 10.9998L12 13.9998L22 3.99976"
+            stroke="#226626"
+            strokeWidth="2"
+            strokeLinecap="round"
+            strokeLinejoin="round"
+          />
+        </svg>
```
ravindrasinghshah, line 84: IMO - its always a good idea to keep asset outside of a component. Maybe we can have svg component which can export these svgs to make it more re-usable.

apps/ideaspace/src/components/common/SubmissionAlert/Alert.js
```diff
@@ -0,0 +1,183 @@
...
+          from {
+            opacity: 0;
+            transform: translateY(-20px);
+          }
+          to {
+            opacity: 1;
+            transform: translateY(0);
+          }
+        }
+      `}</style>
```
ravindrasinghshah, line 143: This animation seems like a re-usable code too - maybe its a good idea to take it out of component in common css.

apps/ideaspace/src/components/modules/SubmissionForm/SubmissionForm.js
```diff
@@ -94,15 +96,16 @@ function SubmissionForm() {
       const data = cleanData(await agent.Ideas.post(values));
 
       if (data.ideaName) {
+        isProgrammaticNavigation.current = true;
         setunsavedChanges(false);
         router.push(`workshop/${data.id}`);
       } else {
-        alert('Unable to register your idea.');
+        throw new Error('Unable to register your idea.');
```
ravindrasinghshah, line 103: I am not sure if throwing error will be caught by error boundaries - can you please confirm the UI/UX behavior of this?

apps/ideaspace/src/components/common/IdeaForm/IdeaForm.js
```diff
@@ -110,12 +112,22 @@ const IdeaForm = ({
     };
   }, []);
 
-  const handleSubmit = (values, actions) => {
-    submitHandler(values, actions);
-    if (!editMode) {
-      setSuccessMessageVisible(true);
-      actions.resetForm({ values: initialValues });
-      clearLocalStorage();
+  const handleSubmit = async (values, actions) => {
```
ravindrasinghshah, line 115: Small nit: Its a good practice to suffix method with Async keyword - it helps improve readability of code. `handleSubmitAsync`

apps/ideaspace/src/components/common/SubmissionAlert/Alert.js
```diff
@@ -0,0 +1,183 @@
...
+        return (
+          <>
+            <strong>Failed to post your idea.</strong> Please try again later.
+          </>
+        );
+      } else {
+        return (
+          <>
+            <strong>Failed to save changes.</strong> Please try again later.
+          </>
```
ravindrasinghshah, line 61: I am not sure how much we want to make this component re-usable but I think we can re-use it for all alert requirements in ideaspace - show alert on idea create, edit and delete. Maybe we should consider passing text as a prop to component to make it more reusable.

apps/ideaspace/src/components/modules/SubmissionForm/SubmissionForm.js
dbradham: We tested this today and it is stable. The code here can actually not execute with the current implementation in `agent.ts`. This is fixed on Abdullah's branch, but we are good to move forward with this branch as is.

apps/ideaspace/src/components/modules/SubmissionForm/SubmissionForm.js
```diff
@@ -91,18 +93,20 @@ function SubmissionForm() {
     setSending(true);
 
     try {
-      const data = cleanData(await agent.Ideas.post(values));
-
-      if (data.ideaName) {
+      const response = await agent.Ideas.post(values);
+      if (response.error) {
+        console.log('heree');
```
dbradham, line 98: @Abd-pathan can we remove this console.log?

Abd-pathan: Sure.

PR #2625, opened 2025-07-24
fix(notification avatar null field): add optional src alt for avatar …

Ahsaask:
- quick fix for avatar null field issue

### DIFF
--- a/packages/UI/src/components/NotificationItem/NotificationItem.tsx
+++ b/packages/UI/src/components/NotificationItem/NotificationItem.tsx
@@ -57,7 +57,7 @@ interface NotificationProps extends VariantProps<typeof notificationStyles> {
   targetLink: string;
   timeStamp: string;
   action: NotificationEvent['data']['attributes']['action'];
-  avatar: { src: string; alt: string };
+  avatar?: { src: string; alt: string };
   className?: string;
 }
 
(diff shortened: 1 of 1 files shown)


packages/UI/src/components/NotificationItem/NotificationItem.tsx
```diff
@@ -110,8 +110,8 @@ function NotificationItem({
       >
         <div className={styles.avatarContainer()}>
           <Avatar
-            src={avatar.src}
-            alt={avatar.alt}
+            src={avatar?.src || ''}
+            alt={avatar?.alt || name || 'User avatar'}
```
j-peace, line 114: The alt looks pretty nice!
Really defensive development!

packages/UI/src/components/NotificationItem/NotificationItem.tsx
```diff
@@ -110,8 +110,8 @@ function NotificationItem({
       >
         <div className={styles.avatarContainer()}>
           <Avatar
-            src={avatar.src}
-            alt={avatar.alt}
+            src={avatar?.src || ''}
```
j-peace, line 113: Just checking what happens when you try to run the front-end with `src={''}` ?
It breaks the screen or just don't show the avatar as expected?

j-peace (changes requested):

packages/UI/src/components/NotificationItem/NotificationItem.tsx
dbradham: This will just show the alt text since the empty string will cause the image not to render as expected.

dbradham (approved):

chungthuang (approved):

PR #2627, opened 2025-07-29
fix(home page buttons): fix home page buttons

dbradham:


packages/UI/src/components/atoms/Button/Button.tsx
```diff
@@ -119,7 +118,6 @@ export default function Button(props: PropsType) {
       : '',
     STYLES.size[safeSize],
     STYLES.iconPosition[safeIconPosition],
-    className || '',
   ].join(' ');
```
j-peace, line 121: Hi @dbradham I realize the we still have `atoms.Button` component with the `className` property at `apps/website/src/pages/resources.js`

Please take a look on that to verify if any changes is required.

dbradham: Great catch, updated!

chungthuang (approved):

PR #2695, opened 2025-10-23
2618 validate unique idea name with clear error messaging

Abd-pathan:
Implemented frontend validation for “Idea name” field to ensure that idea name is unique. When a user enters a name that already exists, a error message is displayed.

Changes Included:
Added validation logic to check for duplicate idea names.
Display inline error message: “This idea name is already in use. Please try something else.”

On form submission, if the idea name is taken: 
1. Scrolls to the “Idea name” field.
2. Displays the same inline error message.
3. Prevents submission until a unique name is entered.


AlinaU1 (approved):
👍

PR #2739, opened 2026-01-25
Skill, Interests Edit Section Fixed

prajuyadv:


apps/user-profile/src/components/modules/UserProfile/EditProfileModal/SettingPage/SettingPage.js
```diff
@@ -51,11 +55,46 @@ function SettingPage({ onClose }) {
...
+                { withCredentials: true }
+              );
+            }
+
+            // Save skills to Strapi user (only if changed)
+            if (editProfileState.changes.skillsChanged) {
+              const userId = userData?.id;
+              const skillIds = (editProfileState.skills || []).map((s) => s.id);
+
+              await axios.put(
```
dbradham: can we make this api call using agent class?

dbradham (approved):
Change looks great! Thanks for spending time to address the feedback and use agent class!

PR #2749, opened 2026-02-02
IdeaSpace: Removed ideaOwner populate from query

ravindrasinghshah:
[Investigate and Improve Page Load Time for Browse Ideas page to Reduce Drop-Offs](https://github.com/dev-launchers/dev-launchers-platform/pull/2749) 
IdeaSpace browse page query - removed ideaOwner populate from the api call. I couldn't find any trace where ideaOwner values are used. Query also fetches author object, which is used to display user name, profile image.


dbradham (approved):
looks good

PR #2834, opened 2026-05-27
fix(user-profile): remove edit profile modal header

liangz25:


liangz25:
I confirmed that the issue was caused by showHeader={true} being passed into the shared Modal component inside EditProfileModal.js.

I removed the header flag from the Edit Profile modal and verified locally that the black Dev Launchers banner no longer appears.

I intentionally avoided modifying the shared ModalHeader component to minimize the risk of affecting other modals that still rely on the existing header behavior.

apps/user-profile/src/components/modules/UserProfile/EditProfileModal/EditProfileModal.js
dbradham, line 6: we may want to get rid of the `showHeader` functionality within the Modal itself.

dbradham (approved):
looks good to meet the requirements of the ticket, but left a comment worth reviewing

apps/user-profile/src/components/modules/UserProfile/EditProfileModal/EditProfileModal.js
dbradham: After further discussion with @matthewmendiola this is worth a quick follow up ticket to remove this functionality from the Modal.

PR #2837, opened 2026-05-29
fix(navigation): removed the debugging statement from navigation

royshunhung:
remove debugging line from navigation.
added or statement to use default comment for historical comment

This is followup from pull request 2835 to remove the debugging statement. I also added the logic for historical comment that was missed.

Before
image

After
image


dbradham (approved):
looks good

Issue #2383, opened 2025-05-13
Allow users edit their comments in the comment feed


AlinaU1:
Enable users to edit their own comments directly within the comment feed.

**Acceptance Criteria:**
- [ ] A button or an icon is displayed to edit a comment
- [ ] Users can edit only their own comments.
- [ ] An “Edit” option is available on each comment a user has authored.
- [ ] After editing, the comment updates in place.
- [ ] Optionally, display a small “(edited)” label next to updated comments.

**Additional notes / optional:** 
- When editing, the comment turns into an editable text field.
- Changes can be saved or cancelled.
- Ensure validation still applies (e.g., character limit, empty content).
- Editing should not affect the comment order unless designed to do so.

royshunhung:
Backend functionality completed for API call. Currently. working on front end button setup

royshunhung:
PR merged with main. Closing

PR #2847, opened 2026-07-26
feat(comment): setup comment change for user to be able to update or delete own comments

royshunhung:
Update front end connection to backend for comment owner check.
Also setup new button for user to delete own comment.
Add logic to prompt for delete comment

feature #2383

image


apps/ideaspace/src/components/modules/WorkshoppingPage/CommentsForm/SingleComment.js
```diff
@@ -69,6 +76,63 @@ function SingleCommentComponent(props) {
...
+  }
+
+  function handleCancelEdit() {
+    setEditText(props.children);
+    setIsEditing(false);
+  }
+
+  async function handleSaveEdit() {
+    const trimmed = editText.trim();
+    if (!trimmed || trimmed === props.children) {
```
chungthuang, line 115: What does `trimmed === props.children` mean?

apps/ideaspace/src/components/modules/WorkshoppingPage/CommentsForm/SingleComment.js
```diff
@@ -69,6 +76,63 @@ function SingleCommentComponent(props) {
...
+      setIsEditing(false);
+      return;
+    }
+    setIsSaving(true);
+    try {
+      await agent.Comments.put(props.id, { data: { text: trimmed } });
+      props.onCommentUpdated?.(props.id, trimmed);
```
chungthuang, line 123: What does this do?

apps/ideaspace/src/components/modules/WorkshoppingPage/CommentsForm/SingleComment.js
```diff
@@ -69,6 +76,63 @@ function SingleCommentComponent(props) {
...
+
+  async function handleDeleteClick() {
+    if (await confirmDelete()) {
+      setIsDeleting(true);
+      try {
+        await agent.Comments.delete(props.id);
+        props.onCommentDeleted?.(props.id);
+      } catch (error) {
+        console.error(error);
+        setIsDeleting(false);
```
chungthuang, line 141: Do we want to put this to the final branch?

chungthuang (approved):

apps/ideaspace/src/components/modules/WorkshoppingPage/CommentsForm/SingleComment.js
royshunhung: We have this here to filter out empty comment when it is just empty space

royshunhung: that part is just checking if the edited message is the same as original comment

apps/ideaspace/src/components/modules/WorkshoppingPage/CommentsForm/SingleComment.js
royshunhung: The first line is to prevent double saving. 

This is for making the network request and also making sure the change persists on the UI

apps/ideaspace/src/components/modules/WorkshoppingPage/CommentsForm/SingleComment.js
royshunhung: Had this remove to cascade the error to front end for user to know the change has not been made

